PR #12, opened 2018-01-19
Alert banner messaging

CourtneyLO:


bower.json
```diff
@@ -35,7 +35,8 @@
     "o-colors": "^4.1.5",
     "o-typography": "^5.4.2",
     "o-share": "^6.0.0",
-    "o-grid": "^4.3.6"
+    "o-grid": "^4.3.6",
+		"n-alert-banner": "^0.3.0"
```
tavvy: indentation

src/client/bottom-slot.js
```diff
@@ -1,3 +1,4 @@
+import imperativeOptions from './lib/imperative-options';
```
tavvy: The config for these two message types (o-banner and alert-banner) may diverge over time, and so I think its better that we keep the `imperativeOptions` functions separate. The main purpose is to clearly define what the options are and how they are set instead of a basic object.assign

src/client/top-slot.js
```diff
@@ -0,0 +1,41 @@
+import imperativeOptions from './lib/imperative-options';
+
+const nAlertBanner = require('n-alert-banner');
+const {	messageEvent, listen } = require('./utils');
```
tavvy: spacing

src/components/remove-close-button.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function removeCloseButton (alertBanner, done) {
```
tavvy: I think it would be neater to build this into `n-alert-banner` as a config option. either something like `data-n-alert-no-close` on the markup or `{noClose: true}` when initing.

CourtneyLO: I have built this in alertBanner already, it is set to closeButton with the default being true (display close button). I just wasn't sure how to include it when it is a declarativeElement. I will fix this on Monday

src/client/bottom-slot.js
```diff
@@ -27,7 +28,9 @@ module.exports = function ({ config={}, guruResult, customSetup }={}) {
 	if (actions.length === 0) {
 		// if no actions specified in markup then default to adding it to the
 		// button element (this can happen when declared imperatively)
-		actions = banner.innerElement.querySelectorAll(BANNER_BUTTON_SELECTOR);
+		let buttonAction = banner.innerElement.querySelectorAll(BANNER_BUTTON_SELECTOR);
+		let linkAction = banner.innerElement.querySelectorAll(BANNER_LINK_SELECTOR);
+		actions = buttonAction ? buttonAction : linkAction;
```
tavvy: this can be `actions = buttonAction || linkAction`

src/client/top-slot.js
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	// attach event handlers
+	let actions = alertBanner.innerElement.querySelectorAll(ALERT_ACTION_SELECTOR);
+	if (actions.length === 0) {
+		// if no actions specified in markup then default to adding it to the
+		// button element (this can happen when declared imperatively)
+		let buttonAction = alertBanner.innerElement.querySelectorAll(ALERT_BANNER_BUTTON_SELECTOR);
+		let linkAction = alertBanner.innerElement.querySelectorAll(ALERT_BANNER_LINK_SELECTOR);
+		actions = buttonAction ? buttonAction : linkAction;
```
tavvy: same as other comment

src/client/top-slot.js
```diff
@@ -0,0 +1,65 @@
...
+	}
+	listen(alertBanner.alertBannerElement, 'n.alertBannerClosed', generateEvent('close'));
+	listen(alertBanner.alertBannerElement, 'n.alertBannerOpen', generateEvent('view'));
+	if (actions && actions.length > 0) {
+		actions.forEach((el) => { listen(el, 'click', generateEvent('act')); });
+	}
+
+	//show alertBanner
+	if (customSetup) {
+		customSetup(alertBanner, () => { alertBanner.open(); });
```
tavvy: just thought we will also want to copy over
```
		customSetup(banner, ({ skip=false }={}) => {
 			if (skip) {		 			if (skip) {
 				document.body.dispatchEvent(generateEvent('skip'));		 				document.body.dispatchEvent(generateEvent('skip'));
 			} else {		 			} else {
 				banner.open();		 				banner.open();
 			}		 			}
 		});		 		});
``` 
from `bottom-slot`

PR #20, opened 2018-02-01
add anon subscribe now banner

tavvy:
Porting the anonymous subscribe now banner to the n-messaging-client. In the core experience only the button (a tag) is clickable, this is enhanced to widen the click area to the whole background image for JS users.

| original | n-messaging |
| --- | --- |
| screen shot 2018-02-02 at 16 07 53 | screen shot 2018-02-02 at 16 19 22 |

| Chrome | Firefox | IE | Edge | Safari |
| --- | --- | --- | --- | --- |
| Enhanced | Enhanced | Core* | Enhanced | Enhanced|

\* Seems to be an JS issue with the demo page which stops the banner from being enhanced, haven't dug into it but the navigation is also broken.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -36,7 +36,7 @@
     "o-typography": "^5.4.2",
     "o-share": "^6.0.0",
     "o-grid": "^4.3.6",
-    "n-alert-banner": "^0.7.0",
+    "n-alert-banner": "^0.8.0",
     "o-icons": "^5.0.0"
   },
   "devDependencies": {
--- a/demos/templates/index.html
+++ b/demos/templates/index.html
@@ -1,3 +1,5 @@
+<div class="o-header-before"></div>
+
 <div id='site-navigation'>
 	{{> o-header }}
 </div>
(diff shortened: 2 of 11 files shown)


tavvy:
Couple of things we will need to add to the markup to make sure we mimic / don't break existing tracking:
- On the containing div: `data-trackable="marketing-promo:header"`
- On the larger "hit" area: `data-trackable="marketing-promo:box"`
- On the link itself: `data-trackable="marketing-promo:link"`

This will be annoying, so I will check where in the business we actually use these metrics. If it is in two places then it makes sense to just ask people to change their metric queries.

tavvy:
👍 check about migrating the `data-trackable` and we do not need to worry about that as part of this.

liamkeaton:
@tavvy added support for supplying custom ```data-trackable``` values to the template partials, this can always be reverted if business no longer need these.

tavvy:
👍 lgtm

PR #21, opened 2018-02-02
use n-ui-foundations

tavvy:
~related: https://github.com/Financial-Times/n-alert-banner/pull/20~
~related: https://github.com/Financial-Times/o-banner/pull/38 going to give the same treatment to `o-banner` and see if origami are happy with it or not.~

The tweaks will ensure that messages render correctly, even if `n-ui-foundations` is not explicitly imported via the app in which `n-messaging-client` is included.

 🐿 v2.6.0

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -1,17 +1,33 @@
+@import "o-buttons/main";
 @import "o-colors/main";
+@import "o-grid/main";
 @import "o-icons/main";
 @import "o-typography/main";
-@import "o-buttons/main";
 
-@import 'o-banner/main';
 @import 'n-alert-banner/main';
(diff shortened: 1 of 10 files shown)


main.scss
```diff
@@ -1,3 +1,8 @@
+@import "o-colors/main";
```
wheresrhys: could also just import n-ui-foundations instead of these... I think

tavvy: the issue with that is that `n-ui-foundations` is not silent, I thought it would be better to explicitly import the things we need, but use `n-ui-foundations` to help resolve conflicts in versions

PR #24, opened 2018-02-05
Writing inital tests for nMessagingPresenter

liamkeaton:
Wanted to get thoughts on -

1. Completeness of tests
2. Having tests close to code rather than in a test folder

 🐿 v2.7.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ test: verify
 	make smoke
 
 unit-test:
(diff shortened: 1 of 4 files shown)


liamkeaton:
@tavvy sure, just wanted to float the idea as it can be clearer which tests relate to which files when they sit along side. I will move to a centralised ```/test``` folder.

PR #27, opened 2018-02-07
Add lionel slider

remybach:
A lot of this was ported directly from [Financial-Times/n-marketing](https://github.com/Financial-Times/n-marketing) with the only changes being around separating out the template and updating the values dynamically.

local ft com_5005_

🐿 v2.7.0

### DIFF
--- a/src/client/bottom-slot.js
+++ b/src/client/bottom-slot.js
@@ -6,7 +6,7 @@ const BANNER_ACTION_SELECTOR = '[data-n-messaging-banner-action]';
 const BANNER_BUTTON_SELECTOR = '.n-messaging-banner__button';
 const BANNER_LINK_SELECTOR = '.n-messaging-banner__link';
 
-module.exports = function ({ config={}, guruResult, customSetup }={}) {
+module.exports = function (flags, { config={}, guruResult, customSetup }={}) {
 	let banner;
 	const generateEvent = config.id && messageEvent({ messageId: config.id, position: config.position, variant: config.name, flag: config.flag });
 	const declarativeElement = !config.lazy && config.content;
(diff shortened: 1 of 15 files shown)


src/client/bottom-slot.js
```diff
@@ -6,7 +6,7 @@ const BANNER_ACTION_SELECTOR = '[data-n-messaging-banner-action]';
 const BANNER_BUTTON_SELECTOR = '.n-messaging-banner__button';
 const BANNER_LINK_SELECTOR = '.n-messaging-banner__link';
 
-module.exports = function ({ config={}, guruResult, customSetup }={}) {
+module.exports = function (flags, { config={}, guruResult, customSetup }={}) {
```
alexnaish: This makes this a breaking change, if we passed in flags as an additional option it would be backwards compatible?

src/components/marketing/popup-prompt/utils.js
```diff
@@ -0,0 +1,40 @@
+export const padLeft = (string, length, character = '0') => {
```
alexnaish: A neat padLeft just for info:

```
leftPad = (str, length, character = '0') = {
    return `${character.repeat(Math.max(length - str.length,0))}${str}`;
}
```

src/client/bottom-slot.js
remybach: It won't be breaking since it will come through to here as `null` from [src/client/index.js](https://github.com/Financial-Times/n-messaging-client/blob/rb-lionel/src/client/index.js#L10)

remybach:
Closing this for now after discussing with @tavvy a better way of implementing this.

PR #34, opened 2018-02-12
Initial registration notice

remybach:
🐿 v2.7.0


tavvy:
oh and it would point to ft.com/profile

remybach:
Good shout... that looks much better 👍

PR #37, opened 2018-03-06
Add more ad blocking logic

jkerr321:
The ad blocker is now shown via `n-messaging-client` but monitoring is showing that it's being shown more often than the previous implementation. Adding in a bit more logic to better reflect the logic in the previous implementation.

### DIFF
--- a/src/components/ad-blocker-sub-banner/main.js
+++ b/src/components/ad-blocker-sub-banner/main.js
@@ -1,7 +1,6 @@
 const { getCurrentLayout } = require('o-grid');
 
 module.exports = function customSetup (banner, done) {
-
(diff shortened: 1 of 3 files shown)


src/components/ad-blocker-sub-banner/main.js
```diff
@@ -0,0 +1,18 @@
+const { getCurrentLayout } = require('o-grid');
```
remybach: This file looks identical to the one above... any way this could be DRYed up?

tavvy: yep they are we thought about it. but for consistency in how others are set up we decided to duplicate for each.

PR #38, opened 2018-03-08
More ad blocking logic

jkerr321:
it works! 😀 

updated demo to mimic `n-counter-add-blocking` logic (which adds '`isAdBlocking` class to `html`) and logic to look for it

### DIFF
--- a/demos/app.js
+++ b/demos/app.js
@@ -14,6 +14,7 @@ const app = module.exports = express({
 	withNavigation: false,
 	withAnonMiddleware: false,
 	hasHeadCss: false,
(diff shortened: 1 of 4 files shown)


liamkeaton:
Could we not use `n-counter-add-blocking` in the demo?

jkerr321:
@liamkeaton `n-counter-ad-blocking` detects whether the user has an ad-blocker then adds a class to the html. I _think_ the reason we don't use it here is because then the user would actually have to have an ad-blocker on for that class to be added and for our demo to work.

PR #39, opened 2018-03-13
Support core experience; add GDPR consent message

jonnywyatt:
image

Enabled by setting messageSlotTop toggle to 'GdprConsent'

 🐿 v2.8.0

### DIFF
--- a/manifest.js
+++ b/manifest.js
@@ -38,7 +38,6 @@ module.exports = {
 	},
 	gdprConsent: {
 		partial: 'top/gdpr-consent',
-		messageId: 'gdprConsent',
-		coreExperience: true
+		messageId: 'gdprConsent'
 	}
 };
--- a/manifest.js
+++ b/manifest.js
@@ -35,5 +35,10 @@ module.exports = {
 	adBlockerRegBanner: {
 		partial: 'bottom/ad-blocker-reg-banner',
 		messageId: 'adBlockerRegBanner'
+	},
+	gdprConsent: {
+		partial: 'top/gdpr-consent',
+		messageId: 'gdprConsent',
+		coreExperience: true
 	}
 };
--- a/templates/partials/top/gdpr-consent.html
+++ b/templates/partials/top/gdpr-consent.html
@@ -4,6 +4,6 @@
 	contentShort='<p>Stay up to date with the FT.</p>'
 	buttonUrl='https://www.ft.com/preferences/reconsent'
 	buttonLabel='Change your preferences now'
-	appendToElement='#site-navigation'
 	closeButton='true'
+	renderClosed=true
 }}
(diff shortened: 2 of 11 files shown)


templates/partials/top/gdpr-consent.html
```diff
@@ -0,0 +1,9 @@
+{{> n-messaging-client/templates/components/n-alert-banner
+	theme='neutral'
+	contentLong='<p>We\'re improving how you stay up to date with the FT.</p>'
+	contentShort='<p>Stay up to date with the FT.</p>'
+	buttonUrl='https://www.ft.com/preferences/reconsent'
+	buttonLabel='Change your preferences now'
+	appendToElement='#site-navigation'
```
tavvy: I wonder how this works in core mode.`appendToElement` is picked up by the client side script and moves the `element` to the correct place in the dom. For core mode you would need to make sure that the template is imported in the correct place in the dom (i believe this is happening already)

manifest.js
```diff
@@ -35,5 +35,10 @@ module.exports = {
 	adBlockerRegBanner: {
 		partial: 'bottom/ad-blocker-reg-banner',
 		messageId: 'adBlockerRegBanner'
+	},
+	gdprConsent: {
+		partial: 'top/gdpr-consent',
+		messageId: 'gdprConsent',
+		coreExperience: true
```
tavvy: To avoid having "magic" config that pulls things in from elsewhere, would it not be possible to simply add to `templates/partials/top/gdpr-consent.html` a `coreExperience=true` variable. Which is picked up by `templates/components/n-alert-banner.html` accordingly?

tavvy: 🤔 reason that may not work is due to having `class="n-ui-hide"` on the main `templates/slot.html` template

templates/partials/top/gdpr-consent.html
jonnywyatt: Yep, verified it in core mode

manifest.js
jonnywyatt: ah.. I thought you were suggesting another config key earlier, similar to the lazy loading one?

demos/templates/index.html
```diff
@@ -2,9 +2,9 @@
 
 <div id='site-navigation'>
 	{{> o-header }}
+	{{> n-messaging-client/templates/slot type='top' }}
 </div>
 
-{{> n-messaging-client/templates/slot type='top' }}
```
tavvy: Does this need to be in `site-navigation` I don't think this will properly replicate how it looks in the actual apps e.g`next-article`

templates/partials/top/gdpr-consent.html
```diff
@@ -0,0 +1,9 @@
+{{> n-messaging-client/templates/components/n-alert-banner
+	theme='neutral'
+	contentLong='<p>We\'re improving how you stay up to date with the FT.</p>'
+	contentShort='<p>Stay up to date with the FT.</p>'
+	buttonUrl='https://www.ft.com/preferences/reconsent'
+	buttonLabel='Change your preferences now'
+	closeButton='true'
+	coreExperience=true
```
liamkeaton: I feel this shouldn't be called `coreExperience` but be named something like `renderClosed`.

`coreExperience` is business level terminology which could become confusing.

jonnywyatt: done

PR #40, opened 2018-03-14
Add an alternative, much smaller main SASS include

jonnywyatt:
This just contains the CSS needed by those message types that can be server-rendered (currently only 'gdprConsent' type
In the consuming app, @import 'n-messaging-client/critical' into the head stylesheet, so it's inlined. Then @import 'n-messaging-client/main' outside the head stylesheet so the full styles are lazily loaded 

This stylesheet outputs only 7Kb of uncompressed CSS vs 47Kb from the existing main stylesheet

 🐿 v2.8.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,17 @@ Now you can inject the message "slot" template in the relevant place in your mar
 {{> n-messaging-client/templates/slot type='bottom'}}
 ```
 
-Import `n-messaging-client`'s styles to your main css entry:
+Import `n-messaging-client`'s styles to your main css entry. 
+
+If you're using a message type that is server-rendered, import the critical stylesheet into the 'head' section of your main.scss.
+
+```scss
+/* critical.scss */
+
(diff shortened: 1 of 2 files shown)


liamkeaton:
I am by far an expert in this stuff, but should the declarations in the `critical` css be removed from the `main`?

jonnywyatt:
@liamkeaton as the main stylesheet is deferred, I don't think it's harmful to leave them in for the moment. Then an app can choose to include critical, main or both.
The better medium-term solution could be to make the main CSS smaller. There's a lot of duplication, which I know gzip's well, but would still increase the front page's gzipped size by 16% (43Kb -> 50Kb) which seems a lot for a non-core component, given all the recent efforts to make the front page faster?

liamkeaton:
Good point we wouldn't want to break the existing implementations.

I wouldn't know where to start trying to minimise the CSS, seems we are mainly importing o components.

PR #41, opened 2018-03-28
Support tooltips; show fallback message for core users

jonnywyatt:
- for core users, show message bar (server-rendered)
image
- for enhanced users on viewports < M, show message bar with close button
image
- enhanced users on wider viewports get the tooltip only
image

 🐿 v2.8.0

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -34,3 +34,4 @@
 @import './src/components/desktop-app-banner/main';
 @import './src/components/anon-subscribe-now-teal/main';
 @import './src/components/ad-blocker-reg-banner/main';
+@import './src/components/nav-messaging-tooltip';
--- a/src/components/nav-messaging-tooltip.scss
+++ b/src/components/nav-messaging-tooltip.scss
@@ -1,0 +1,41 @@
+.n-alert-banner--neutral {
+	.n-alert-banner__outer {
+		background-color: oColorsGetPaletteColor($name: 'wheat');
+	}
+
+	.n-alert-banner__inner {
+		justify-content: space-between;
+	}
+
+	.n-alert-banner__content {
+		@include oTypographySans($scale: 0);
+		padding-left: 0;
+		padding-right: 0;
+		margin-left: 0;
+		&:before {
+			display: none;
+		}
+		@include oGridRespondTo(M) {
+			padding-right: 24px;
+		}
+	}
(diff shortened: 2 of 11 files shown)


src/components/nav-messaging-tooltip.scss
```diff
@@ -0,0 +1,41 @@
...
+		background-color: oColorsGetPaletteColor($name: 'wheat');
+	}
+
+	.n-alert-banner__inner {
+		justify-content: space-between;
+	}
+
+	.n-alert-banner__content {
+		@include oTypographySans($scale: 0);
+		padding-left: 0;
```
doramatadora: Minor and maybe doesn't apply: can you use shorthand here?

main.scss
```diff
@@ -34,3 +34,4 @@
 @import './src/components/desktop-app-banner/main';
 @import './src/components/anon-subscribe-now-teal/main';
 @import './src/components/ad-blocker-reg-banner/main';
+@import 'src/components/nav-messaging-tooltip';
```
doramatadora: minor: `./` for consistency

src/components/nav-messaging-tooltip.scss
jonnywyatt: Do you mean for the padding?

src/components/nav-myft/main.js
```diff
@@ -0,0 +1,17 @@
...
+	const contentEl = banner.innerElement.querySelector('.n-alert-banner__content');
+	const targetElement = document.querySelector('.o-header__top-link--myft').parentElement;
+	const opts = {
+		target: 'o-header__top-link--myft',
+		content: contentEl.innerHTML,
+		showOnConstruction: true,
+		position: 'below'
+	};
+
+	const tooltip = new oTooltip(targetElement, opts);
```
tavvy: More a question, and perhaps my mis-understanding of the docs. But should the first parameter passed into `oTooltip` be the tooltip element? and not the target? https://github.com/Financial-Times/o-tooltip#constructing-an-o-tooltip

jonnywyatt: That's if the tooltip is already rendered out (declaratively) rather than you want to construct it dynamically (or imperatively as the doc calls it). I think.

PR #43, opened 2018-04-16
Add daily-digest tooltip message

dan-searle:
More to do...

### DIFF
--- a/templates/partials/top/daily-digest.html
+++ b/templates/partials/top/daily-digest.html
@@ -1,0 +1,7 @@
+{{> n-messaging-client/templates/components/n-alert-banner
+theme='neutral'
+contentLong='<p>Click here for your daily digest.</p>'
+contentShort='<p>Click here for your daily digest.</p>'
+closeButton='true'
+renderOpen=true
+}}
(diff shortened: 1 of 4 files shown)


templates/partials/top/daily-digest.html
```diff
@@ -0,0 +1,7 @@
+{{> n-messaging-client/templates/components/n-alert-banner
+theme='neutral'
+contentLong='<p>Click here for your daily digest.</p>'
```
jonnywyatt: The tooltip is on wider viewports only - on mobile it becomes a messaging bar. So the text needs to make sense in that context

dan-searle: I see, thanks. `contentLong` is for the messaging bar, and `contentShort` is for the tooltip?

jonnywyatt: contentShort is shown up to and including S viewport widths, and contentLong beyond that
CSS is in [n-alert-banner](https://github.com/Financial-Times/n-alert-banner/blob/master/src/scss/alert-banner/_alert-banner-mixins.scss#L42-L73)
So yes if tooltips are enabled, they'll use contentLong

dan-searle: Got it, thanks!

dan-searle:
This is parked for now. We're going to implement temporary solution with [n-myft-ui](https://github.com/Financial-Times/n-myft-ui) for the A-B test, and revisit a better solution later.

PR #44, opened 2018-04-19
n-alert-banner to o-message

CourtneyLO:


critical.scss
```diff
@@ -1,22 +1,19 @@
...
 @include oFontsInclude(MetricWeb);
 @include oFontsInclude(MetricWeb, $weight: semibold);
 
-@include nAlertBanner($class: 'n-alert-banner', $themes: 'neutral');
-
-// ensure the top n-alert-banner inner content is aligned to page content
-.n-alert-banner__inner {
-	@include oGridContainer();
-}
+@include oMessage($class: 'n-alert-banner', $types: 'alert', $status: 'all');
```
jonnywyatt: Do we know how much CSS this will bring in? It will be inlined into front / stream / article pages

templates/components/n-alert-banner.html
```diff
@@ -1,30 +1,26 @@
-<div class="n-alert-banner {{#unless renderOpen}}n-alert-banner--closed{{/unless}} n-alert-banner--{{theme}}{{#if customClass}} {{customClass}}{{/if}}" data-n-component="n-alert-banner" data-n-messaging-component="" data-n-alert-banner-auto-open="false" data-n-alert-banner-close-button="{{#if closeButton}}{{closeButton}}{{else}}true{{/if}}" {{#if customDataTrackableBanner}}data-trackable="{{customDataTrackableBanner}}"{{/if}}>
-	<div class="n-alert-banner__outer">
-		<div class="n-alert-banner__inner" data-n-alert-banner-inner="">
-			{{#if contentLong}}
-			<!-- Content to display on larger screens -->
-				<div class="n-alert-banner__content n-alert-banner__content--long">
-					{{{contentLong}}}
+<div class="n-alert-banner n-alert-banner--alert-bleed {{#unless close}}close-button-inner{{/unless}} {{#unless renderOpen}}n-alert-banner--closed{{/unless}} n-alert-banner--{{theme}}{{#if customClass}} {{customClass}}{{/if}}" data-n-messaging-component="" data-n-alert-banner-auto-open="false" {{#if customDataTrackableBanner}}data-trackable="{{customDataTrackableBanner}}"{{/if}}>
```
jonnywyatt: Just checking, if this PR removes n-alert-banner, why is the component template remaining?

templates/partials/top/nav-contact-preferences.html
```diff
@@ -1,7 +1,7 @@
 {{> n-messaging-client/templates/components/n-alert-banner
 theme='neutral'
-contentLong='<p><strong>Emails & Alerts</strong> are now part of <strong>Contact preferences</strong>.</p>'
-contentShort='<p><strong>Emails & Alerts</strong> are in <strong>Contact preferences</strong>.</p>'
-closeButton='true'
+contentTitle='Emails & Alerts'
```
jonnywyatt: Is a shorter version for mobile supported? This will wrap or clip at 320px I think

demos/templates/partials/o-header.html
```diff
@@ -92,6 +94,9 @@
 				<li class="o-header__nav-item">
 					<a class="o-header__nav-link o-header__nav-link--primary" href="">Life &amp; Arts</a>
 				</li>
+				<li class="o-header__nav-item" data-id="subnav-alerts" id="[data-id=&quot;subnav-alerts&quot;]" data-content-id="subnav-alerts">
```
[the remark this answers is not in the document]
CourtneyLO: Yeah, I wanted to be able to see the NavContactPreferences tooltip

main.scss
```diff
@@ -1,37 +1,38 @@
+$o-message-is-silent: false;
```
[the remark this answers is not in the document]
CourtneyLO: The o-message docs recommend it be set to false, should it not be?

templates/components/n-alert-banner.html
CourtneyLO: We are just using our own class name instead on o-message so the banner won't be affected by any overrides.

templates/partials/top/nav-contact-preferences.html
CourtneyLO: o-message now wraps content - there is no longer a short and long version

demos/templates/partials/o-header.html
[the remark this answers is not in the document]
jonnywyatt: the data-id attribute should be enough on it's own, it shouldn't need an id as well?

critical.scss
CourtneyLO: It is an origami component so should be. I think it brings in the same amount as n-alert-banner but might be wrong.

demos/templates/partials/o-header.html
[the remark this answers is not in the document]
CourtneyLO: Oh, probably not. I just copied and pasted. I will remove that.

main.scss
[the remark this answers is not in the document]
jonnywyatt: I can't see that recommendation? https://github.com/Financial-Times/o-message#sass

templates/partials/top/nav-contact-preferences.html
jonnywyatt: @draysonandstock are you sure you want this to wrap? On a smaller-screen mobile nearly half the viewport height is taken up with header elements
image

main.scss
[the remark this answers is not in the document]
CourtneyLO: `As with all Origami components, o-message has a silent mode. To use its compiled CSS (rather than using its mixins with your own Sass) set $o-message-is-silent: false; in your Sass before you import the o-message Sass.`

main.scss
```diff
@@ -1,37 +1,38 @@
...
 
-// ensure the top n-alert-banner inner content is aligned to page content
-.n-alert-banner__inner {
-	@include oGridContainer();
-}
-
 // set base font
 .n-alert-banner,
 .n-messaging-banner {
 	font-family: sans-serif;
```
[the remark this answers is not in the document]
CourtneyLO: It does use MetricWeb throughout, to honest, I am not too sure why this is here.

draysonandstock:
@jonnywyatt so although we can always improve the actual message and keep it as short as it required I feel strongly that showing a usable message is better than a shorter unhelpful one. For now would be keen to keep it wrapping but might look at how we can best advise on the actual errors.

main.scss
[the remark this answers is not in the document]
jonnywyatt: It describes silent mode but it's not recommending it? Using silent mode means less CSS is pulled in, which is a bigger deal for the critical CSS than the lazily loaded.
I'm relatively new here but that was my understanding of good practice. It could be worth getting another opinion.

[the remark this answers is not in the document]
CourtneyLO: Yeah, you are right. I am not using the o-message mixins anywhere so I will remove that.

src/components/nav-contact-preferences/main.js
```diff
@@ -13,5 +13,6 @@ module.exports = function customSetup (banner, done) {
 
 	const tooltip = new oTooltip(targetElement, opts);
 	tooltip.tooltipEl.classList.add('n-messaging-client-tooltip');
+	tooltip.tooltipEl.querySelector('.n-alert-banner__content-main').classList.add('n-alert-banner--alert-bleed');
```
jonnywyatt: I can't see this class in n-alert-banner repo, is it needed?

src/components/nav-myft/main.js
```diff
@@ -13,5 +13,6 @@ module.exports = function customSetup (banner, done) {
 
 	const tooltip = new oTooltip(targetElement, opts);
 	tooltip.tooltipEl.classList.add('n-messaging-client-tooltip');
+	tooltip.tooltipEl.querySelector('.n-alert-banner__content-main').classList.add('n-alert-banner--alert-bleed');
```
jonnywyatt: see earlier comment

src/components/nav-contact-preferences/main.js
CourtneyLO: `n-alert-banner__content-highlight` styling will not work without it

main.scss
```diff
@@ -1,35 +1,28 @@
+$o-tooltip-is-silent: false;
```
tavvy: can we not pull this in non-silently?

jonnywyatt: Yes, I can do that. In terms of the quantity of CSS used from that component, I was using pretty much all of it I believe so there was no efficiency gain. However this could still protect us from differing versions of o-tooltip

jonnywyatt: actually it appears this isn't an option - https://financialtimes.slack.com/archives/C02FU5ARJ/p1525171135000006

critical.scss
```diff
@@ -1,44 +1,29 @@
...
 
 @include oFontsInclude(MetricWeb);
 @include oFontsInclude(MetricWeb, $weight: semibold);
 @include oBanner($class: 'n-messaging-banner', $themes: 'all');
-@include nAlertBanner($class: 'n-alert-banner', $themes: 'all');
+@include oMessage($class: 'n-alert-banner', $types: 'alert', $status: 'all');
 
 // ensure the bottom o-banner displays above other elements on page (e.g. imgs on article page)
 .n-messaging-banner {
 	z-index: 100;
```
jonnywyatt: There's a mixin for generating z-indexes, could use that? https://github.com/Financial-Times/n-ui-foundations/blob/master/util/_mixins.scss

src/components/nav-messaging-tooltip.scss
```diff
@@ -1,26 +1,7 @@
...
 
-	.n-alert-banner__outer {
-		@include alertBannerOuter;
+	.n-alert-banner__container {
 		background-color: oColorsGetPaletteColor($name: 'wheat');
-	}
-
-	.n-alert-banner__inner {
-		@include alertBannerInner;
-		justify-content: space-between;
```
jonnywyatt: If this is removed, how is the content spaced?

jonnywyatt:
LGTM 👍

PR #53, opened 2018-05-10
Team trial notification with side bar form

liamkeaton:
screen shot 2018-05-10 at 11 38 16

 🐿 v2.7.0


bower.json
```diff
@@ -36,6 +36,9 @@
     "o-share": "^6.0.0",
     "o-tooltip": "^3.1.1",
     "o-viewport": "^3.1.5",
+    "o-overlay": "^2.4.1",
+    "o-colors": "^4.4.0",
```
CourtneyLO: o-overlay and o-colors are already in the bower-components as dependencies. Is there a reason you need to add them to bower.json?

liamkeaton: I'm using them in my SCSS so I want to ensure they're there for two reasons -

1. If I rely on other components we could break if those components are refactored.
2. It's clear from our bower.json what we are explicitly using.

CourtneyLO: Cool, thanks @liamkeaton

CourtneyLO:
It looks good to me

tavvy:
I am a little worried that the overlay could easily "clash" with the bottom slot messages. At the moment most of those appear on load. So maybe not as much of a concern as the user has to action the top slot banner for the slider to appear. As long as this slider has the highest z index should be fine.

liamkeaton:
Good point, I will check this out and see

liamkeaton:
@tavvy added some z-index'in

screen shot 2018-05-14 at 11 49 02

bower.json
tavvy: One benefit of having these managed by `n-ui-foundations` is you are far less likely to come across version conflicts as most apps depend upon `n-ui` / `n-ui-foundations`

liamkeaton: @tavvy feels a little dangerous to rely on dependencies dependencies as package/version changes would then break us without knowing. Happy to remove them here though if that's the convention?

tavvy: yep I think having read a bit on https://github.com/Financial-Times/n-ui-foundations it seems we should stick to relying on it. `n-ui-foundations` is versioned too, so they shouldn't be able to break things

liamkeaton:
@Financial-Times/ft-com-conversion this is ready to go now!

liamkeaton:
@Financial-Times/ft-com-conversion please review 👍

src/components/team-trial/main.js
```diff
@@ -0,0 +1,41 @@
...
+<iframe class="${iframeClass}" src="${src}"></iframe>`;
+
+module.exports = function customSetup (banner, done) {
+	const overlayTrigger = banner.innerElement.querySelector('.n-alert-banner__action a');
+	const closeTrigger = banner.innerElement.querySelector('.n-alert-banner__action--secondary a');
+
+	if (overlayTrigger) {
+		const src = overlayTrigger.href + (/\?/.test(overlayTrigger.href) ? '&' : '?') + 'embedded=true';
+		const options = {
+			html: `${generateHtml(src)}`,
```
remybach: Don't think you need to use the template literal here.

PR #55, opened 2018-05-14
Flags for href

debugwand:


jonnywyatt:
It's not a huge deal, but when the new cookies page goes live we'll be introducing a redirect from the old one anyway, so this isn't really needed?

debugwand:
in that case I will not merge so that we're not adding a large flags object in the slots

PR #57, opened 2018-05-16
Kb/deprecated tls

ker-an:
To show an error message to users on older TLS versions (v1 and v1.1) before they are deprecated.

### DIFF
--- a/templates/partials/top/tls-deprecation.html
+++ b/templates/partials/top/tls-deprecation.html
@@ -1,8 +1,8 @@
 {{> n-messaging-client/templates/components/n-alert-banner
 	theme='error'
-	contentLong='<p><b>Warning!</b> After 18th June 2018 the web browser you are currently using won’t be able to access FT.com.</p>'
-	contentShort='<p>After 18th June 2018 this browser won’t be able to access FT.com.</p>'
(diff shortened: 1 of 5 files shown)


templates/partials/top/tls-deprecation.html
```diff
@@ -0,0 +1,7 @@
+{{> n-messaging-client/templates/components/n-alert-banner
+	theme='error'
+	contentLong='<p><b>Warning!</b> After 18th June 2018 the web browser you are currently using won’t be able to access FT.com.</p>'
```
tavvy: I think you _may_ need a `contentShort` for smaller screen sizes

adgad: We're kind of struggling to get a meaningful message that fits in one line! Is there any particular reason the banner has a fixed height? Do you reckon we could make it wrap?

tavvy: Not that I'm aware of, I imagine it was a design consideration though. I believe @jonnywyatt has multiple lines in the new privacy policy message

ker-an:
Long:

image

Short:

image

@tavvy happy last day! could you take a look at the PR and add your seal of approval?

PR #62, opened 2018-05-22
Fix FOUC in privacy policy message

jonnywyatt:
Remove privacy policy styles from main stylesheet as already in critical; increase specificity of privacy policy styles

 🐿 v2.8.0

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -36,4 +36,3 @@ $o-tooltip-is-silent : false;
 @import './src/components/desktop-app-banner/main';
 @import './src/components/anon-subscribe-now-teal/main';
 @import './src/components/cookie-consent/main';
-@import './src/components/privacy-policy/main';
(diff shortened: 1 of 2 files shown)


liamkeaton:
Am I going mad or did you do this yesterday?

main.scss
```diff
@@ -36,4 +36,3 @@ $o-tooltip-is-silent : false;
 @import './src/components/desktop-app-banner/main';
 @import './src/components/anon-subscribe-now-teal/main';
 @import './src/components/cookie-consent/main';
-@import './src/components/privacy-policy/main';
```
remybach: Oops... hold on... are you meant to be removing this?

jonnywyatt:
@liamkeaton ha ha you're not going mad. I did remove from main.scss. However the specificity issue meant the styles broke so I had to revert it.
Raising the specificity is a sticking plaster, those styles need to move in to n-alert-banner then it won't be needed. I'll make a tech debt note on our board

main.scss
jonnywyatt: Yes. The issue is, they're also included in critical.scss. When the CSS is built, it de-dupes in favour of the last occurrence of a style. As that second occurrence was in main.scss, that's the one that remains. That stylesheet is lazily loaded, but the markup for the message is server-rendered, hence the FOUC

liamkeaton:
@jonnywyatt awesome, I believe we will be moving to `o-message` soon which may make things easier/different...

PR #63, opened 2018-05-22
Move most styles to critical stylesheet

jonnywyatt:
:(
There was still a minor FOUC on the privacy policy message which I hadn't detected locally.
I've moved most styles to the critical stylesheet. I'd been avoiding this as it increases the amount of CSS being inlined in the pages.
Going forward, it would help to either:
- not server-render messages ie nothing for core experience
- make the n-messaging-client styles as compact as possible - at the moment there's lots of duplication which is making them bigger than necessary

 🐿 v2.8.0


critical.scss
```diff
@@ -6,13 +6,32 @@
 
 @import 'n-alert-banner/main';
 @import 'o-banner/main';
+$o-tooltip-is-silent : false;
```
alexnaish: Do we need to disable silent mode if we're using the mixin on line 16?

jonnywyatt: The oBanner mixin? I don't believe so, no. It's not actually possible I think to use the tooltips in silent mode as references to o- classes are hardcoded in the JS

PR #64, opened 2018-05-22
Add cookie consent endpoint and update copy

debugwand:
🐿 v2.8.0

### DIFF
--- a/src/components/cookie-consent/main.js
+++ b/src/components/cookie-consent/main.js
@@ -16,24 +16,44 @@ module.exports = function customSetup (banner, done) {
 		bannerElem.parentNode.removeChild(bannerElem);
 	};
 
-	const accept = (elem, event) => {
-		const elemHref = elem.getAttribute('href');
-		if (elemHref) { // pause to allow us to save new state
-			event.preventDefault();
-		}
+	const setAccepted = () => {
 		cookieStore.set(LOCAL_STORE_KEY, 'true', {
 			domain: '.ft.com',
 			maxAge: 60*60*24*365*2
 		});
 		removeBanner();
+	};
+
+	const acceptAction = (elem, event) => {
+		const elemHref = elem.getAttribute('href');
+		if (elemHref) { // pause to allow us to save new state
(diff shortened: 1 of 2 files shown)


src/components/cookie-consent/main.js
```diff
@@ -16,24 +16,44 @@ module.exports = function customSetup (banner, done) {
...
+		const elemAction = elem.getAttribute('action');
+		const elemMethod = elem.getAttribute('method');
+		event.preventDefault();
+		// call the consent proxy to set other cookies to yes, but still hide banner if those fail. User can manage via preferences pages.
+		return fetch(elemAction, {
+			method: elemMethod
+		})
+		.then(setAccepted)
+		.catch(error => {
+			return { error };
```
jonnywyatt: Will this get logged out anywhere?

src/components/cookie-consent/main.js
```diff
@@ -16,24 +16,44 @@ module.exports = function customSetup (banner, done) {
...
+		});
+	};
+
 	const setup = () => {
 		const accepted = [].slice.call(wrapper.querySelectorAll('[data-action="accepted"]'));
-		accepted.forEach(elem => elem.addEventListener('click', (event) => accept(elem, event), false));
+		const acceptForm = [].slice.call(wrapper.querySelectorAll('[data-action="accept-form"]'));
```
jonnywyatt: You can call forEach directly on the nodeList results - https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll

debugwand: I initially did this and had to change it, trying to recall why. Will dig through commits

debugwand: yep IE11: https://github.com/Financial-Times/n-messaging-client/commit/29ace9df5a4c7868c2341c98f54a568e3d777ed0

src/components/cookie-consent/main.js
debugwand: o-errors I think is picking this up

PR #69, opened 2018-05-23
Add a flag so we can test ammit whilst not showing banner to users

debugwand:
🐿 v2.8.0


tavvy:
If we just want to validate that ammit is correctly cohorting people there is another way:

1) create a dummy flag `cookieBannerTest` that doesn't do anything
2) add the `cookieConsent` cohort to the flag, make the flag active
3) you should start to see you cohort appearing here: http://grafana.ft.com/d/000000191/next-ammit?refresh=1m&panelId=28&fullscreen&orgId=1 this is caused by the metrics tracking here: https://github.com/Financial-Times/next-ammit-api/blob/master/server/cohorts/cookie-consent.js#L8

tavvy:
Ideally we never "over ride" the decision made by ammit. Otherwise the metrics could be skewed.
I.e: " page views where the messageSlotBottom = cookieBanner" would not actually mean that we did show the banner.

PR #79, opened 2018-06-08
Cookie banner variant c

jonnywyatt:
- Add variant C of banner
- Make variant A use variant B's styling
- fix an issue with the bottom slot / messaging banner where declarative attributes weren't being used by o-banner

image

### DIFF
--- a/manifest.js
+++ b/manifest.js
@@ -19,6 +19,10 @@ module.exports = {
 		partial: 'bottom/cookie-consent',
 		messageId: 'cookieConsentB'
 	},
+	cookieConsentC: {
+		partial: 'bottom/cookie-consent',
+		messageId: 'cookieConsentC'
+	},
 	marketingPopupPrompt: {
 		partial: 'bottom/lazy',
 		messageId: 'marketingPopupPrompt',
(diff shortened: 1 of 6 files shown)


manifest.js
```diff
@@ -13,12 +13,16 @@ module.exports = {
 	},
 	cookieConsentA: {
 		partial: 'bottom/cookie-consent',
-		messageId: 'cookieConsentA'
+		messageId: 'cookieConsentB'
```
liamkeaton: Will this have any impact on the tracking?

jonnywyatt: Yes, intentionally, see [this thread](https://financialtimes.slack.com/archives/C8M61LS5A/p1528455979000248) for explanation of why there are still some users with variant A.
As we're going to make any remaining variant A banners look the same as variant B (there's a monetary value to this as variant B scored much better in the acceptance rate), we might as well have them reported as in the same bucket.

liamkeaton: Ok cool, so essentially `cookieConsentA` is now `cookieConsentB`. Unfortunately  we wont know when to retire `cookieConsentA` as we wont be able to pull any numbers from it?

jonnywyatt: So the decision is, to not reassign variant A to variant B

PR #83, opened 2018-06-13
Improving naming consistency

CourtneyLO:
Changes naming of `b2b upsell banner` to `invite colleagues` and `desktop app banner` to `app promoting banner` for naming consistency. 

 🐿 v2.8.0


manifest.js
```diff
@@ -3,12 +3,12 @@ module.exports = {
 		partial: 'bottom/lazy', // path to template
 		messageId: 'licenceJoinPromptSsiClient' // id as specified in ammit-api
 	},
-	b2bUpsellBanner: {
+	inviteColleagues: {
```
liamkeaton: Is this used as the variant name?

CourtneyLO: There is no variant for this currently and it is not in Ammit. We use to have two invite colleagues banners so that's why it was named this but the other one has been removed

liamkeaton: Cool 👍

PR #89, opened 2018-06-15
Reduce steps

CourtneyLO:
This work attempts to reduce the number of files that need to be changed in order to add a new message. 

- A path ending must be set in the manifest.js. This will then be the path used to get the partial as well as the custom setup. This prevents having to import the file in src/components/index

This work also reduces the amount of unnecessary data that is added to the HTML by importing the manifest.js file into src/client/index.js. 

Still to do: 

I want to get rid of the `variant` in the tracking data as it is the same as the `messageId`. I will need to talk to the data team before doing this as there are tests on the go and therefore this is not in this pr.

This is only the first part of the refactoring, I want to do :)


ft.yml
```diff
@@ -1,3 +1,3 @@
 owner:
-  github: liamkeaton
-  email: liam.keaton@ft.com
+  github: CourtneyLO
```
liamkeaton: https://bukk.it/brbyoink.gif

CourtneyLO: :joy::joy::joy::joy::joy::joy:

src/client/index.js
```diff
@@ -1,21 +1,23 @@
 const fetchAsyncConfig = require('./next-messaging-guru-client');
 const topSlot = require('./top-slot');
 const bottomSlot = require('./bottom-slot');
-const components = require('../components');
+const manifest = require('../../manifest');
+
+const LAZY_REGEXP = /\/lazy$/;
```
remybach: Could the path ever end in a trailing `/`?

src/components/bottom/cookie-consent/main.scss
```diff
@@ -1,7 +1,7 @@
 /* stylelint-disable selector-class-pattern */
-.n-messaging-slot--cookieConsentA,
-.n-messaging-slot--cookieConsentB,
-.n-messaging-slot--cookieConsentC {
+[data-n-messaging-name="cookieConsentA"],
```
liamkeaton: Could this be done in a separate PR?

src/components/top/anon-subscribe-now-teal/main.js
```diff
@@ -1,6 +1,7 @@
 module.exports = function customSetup (banner, done) {
-	const bannerActions = banner.querySelector('n-alert-banner__actions');
-	const bannerButton = banner.querySelector('n-alert-banner__actions__primary');
+
+	const bannerActions = banner.messageElement.querySelector('.n-alert-banner__actions');
```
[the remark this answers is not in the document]
CourtneyLO: Should I create a separate branch for this or just in future?

src/client/index.js
```diff
@@ -36,8 +38,14 @@ module.exports = {
...
-	setupHandler (name, position) {
-		return components[position] && components[position].hasOwnProperty(name) && components[position][name];
+	setupHandler (path) {
+		try {
+			if (path) return require(`../components/${path}/main`);
+		} catch (error) {
+			// Not all variants have a custom setup files and therefore this prevents an error being throw
+			// console.error(error)
+			return;
+		}
```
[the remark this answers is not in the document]
remybach: You could use https://nodejs.org/api/fs.html#fs_fs_existssync_path to check if it exists beforehand?

Other than that, I think you'd be ok with doing it this way. The only thing I'd change is maybe doing `if (!path) return;` before the try/catch even.

src/client/index.js
liamkeaton: I've never liked this way of specifying it's to be lazy loaded. Could we have this as an option in the manifest?

src/client/index.js
```diff
@@ -1,21 +1,23 @@
...
 
 module.exports = {
 	init: function () {
 		const slots = document.querySelectorAll('[data-n-messaging-slot]');
 		const messages = slots && Array.prototype.slice.call(slots).map(elm => {
 			const dataSet = elm.dataset || {};
+			const variant = manifest[dataSet.nMessagingName];
 			return {
-				position: dataSet.nMessagingPosition,
+				position: dataSet.nMessagingSlot,
```
jonnywyatt: could rename the key as 'slot' for consistency?

src/components/top/anon-subscribe-now-teal/main.js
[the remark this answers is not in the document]
liamkeaton: I would prefer to keep this PR to only the refactoring, as it may need to be rolled back if things break

src/client/index.js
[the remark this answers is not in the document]
jonnywyatt: How does the build handle this as you're evaluating at runtime?

src/components/bottom/cookie-consent/main.scss
```diff
@@ -1,7 +1,7 @@
 /* stylelint-disable selector-class-pattern */
```
jonnywyatt: If you're not using classnames any more you don't need this stylelint disable line

src/client/index.js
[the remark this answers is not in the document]
remybach: Doh, good point. I didn't notice this was a client side file.

src/components/bottom/cookie-consent/main.scss
CourtneyLO: I couldn't do this one separately because I removed .n-messaging-slot--cookieConsentA as part of the slot.html refactor so this would have affected the styling.

src/components/top/anon-subscribe-now-teal/main.js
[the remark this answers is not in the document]
CourtneyLO: Cool, I will take it out 👍

src/client/index.js
CourtneyLO: That's a great idea, I will do that.

src/client/index.js
CourtneyLO: Yeah sure

src/components/bottom/cookie-consent/main.scss
CourtneyLO: Cool, missed that, thanks.

PR #94, opened 2018-06-29
For user actions on bottom banner, decorate event with action type

jonnywyatt:
This gives data team the ability to distinguish between the type of user action that occurred on a banner.

 🐿 v2.8.0


src/client/utils.js
```diff
@@ -13,6 +13,8 @@ module.exports = {
 				messaging_flag: flag,
 				messaging_variant: variant
 			};
+			if (trackingAttr)
```
liamkeaton: Could we not default it to a sensible value and put it straight into the detail object?

jonnywyatt: Are you sure it's something you want by default? We don't know what the default action would be, it could be anything. I'm wary of setting keys will null or empty string, it just clutters the object without transmitting anything useful

liamkeaton: Not sure what the data team will use it for so can't say, just wanted to clean up a messy if statement :)

PR #97, opened 2018-07-02
added cpc value as query param

jkerr321:
🐿 v2.9.0

Field populating when inspected (locally via demo)
image

### DIFF
--- a/templates/partials/top/team-trial.html
+++ b/templates/partials/top/team-trial.html
@@ -2,7 +2,7 @@
 	theme='neutral'
 	contentTitle='Extend your trial by 30 days'
 	content='with a team trial'
-	buttonUrl='https://next-b2b-prospect.ft.com/form?marketingName=unmasking'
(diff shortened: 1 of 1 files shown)


templates/partials/top/team-trial.html
```diff
@@ -2,7 +2,7 @@
 	theme='neutral'
 	contentTitle='Extend your trial by 30 days'
 	content='with a team trial'
-	buttonUrl='https://next-b2b-prospect.ft.com/form?marketingName=unmasking'
+	buttonUrl='https://next-b2b-prospect.ft.com/form?marketingName=unmasking?cpccampaign=mid_trial_unmasking_barrier'
```
alexnaish: That needs to be an `&` not another question mark since the query string is already started with `?marketingName=unmasking`

jkerr321: Doh!

PR #110, opened 2018-08-16
Readme update

CourtneyLO:
There is still more work to be done on this but I just really wanted to make this look less like it is under construction :)


CourtneyLO:
Thanks, I will have a look at that!

PR #114, opened 2018-09-06
Awaiting results of the entitlements call

liamkeaton:
I've changed this to use `async` `await` to make it clearer what needs to be awaited.

I believe the main issue was that the function was not returning a promise so the caller moved on. This could of been fixed by adding `return` in front of `swgLoader({ customOnwardJourney: true })`. Async functions always return a promise.

### DIFF
--- a/src/components/bottom/swg-entitlements-prompt/main.js
+++ b/src/components/bottom/swg-entitlements-prompt/main.js
@@ -8,7 +8,7 @@ module.exports = (banner, done) => {
 				swg.init();
 				return swg.checkEntitlements();
 			}).then(entitlements => {
-				if (entitlements && !entitlements.granted && !entitlements.hasEntitlements) {
+				if (entitlements && !entitlements.granted && entitlements.hasEntitlements) {
 					// get cta properties from n-swg and apply them to the banner button
 					const ctaProperties = swg.getEntitledOnwardJourneyProps(entitlements);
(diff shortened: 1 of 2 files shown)


liamkeaton:
Good point, demo is compiling and working fine and is building ok on circle. I think it will be finally compiled in the applications that are using it so if they don't support `async` `await` it will go boom. @jkerr321 did you encounter this?

liamkeaton:
@remybach @jkerr321 turned out that just having an entitlements object was not the correct check, I've changed it so that the user is not `granted` but `hasEntitlements`. Is that right?

src/components/bottom/swg-entitlements-prompt/main.js
```diff
@@ -1,22 +1,24 @@
...
-				banner.open();
-			}
-		});
+		return swgLoader({ customOnwardJourney: true })
+			.then(loaded => {
+				swg = loaded;
+				swg.init();
+				return swg.checkEntitlements();
+			}).then(entitlements => {
+				if (entitlements && !entitlements.granted && !entitlements.hasEntitlements) {
```
remybach: I think the logic should be inverted here... we want to show the user this message if they *do* have entitlements.

liamkeaton: Whoops that was me leaving some debug in :P

PR #117, opened 2018-09-18
dynamically generate list of messages for testing

jkerr321:
🐿 v2.10.3

Whilst yak-shaving on another issue on this repo yesterday i noticed the list of features for pa11y testing was hardcoded. As this list changes pretty frequently and the tests are somewhat tucked away (i.e. easy to forget to update) I thought it would be better to generate the list dynamically from the manifest which _should_ be representative of reality

### DIFF
--- a/test/smoke.js
+++ b/test/smoke.js
@@ -1,21 +1,18 @@
 const manifest = require('../manifest');
 
-let messages = [];
-
 const parseSlot = (path) => {
 	if (path.includes('bottom')) return 'messageSlotBottom';
 	if (path.includes('top')) return 'messageSlotTop';
 };
 
 // create array of feature names and slots based on manifest
-for (let key in manifest) {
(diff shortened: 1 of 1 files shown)


test/smoke.js
```diff
@@ -1,29 +1,18 @@
...
-];
+const manifest = require('../manifest');
+
+const parseSlot = (path) => {
+	if (path.includes('bottom')) return 'messageSlotBottom';
+	if (path.includes('top')) return 'messageSlotTop';
+};
+
+// create array of feature names and slots based on manifest
+const createMessage = function (key) {
```
liamkeaton: Oh your gonna hate me but as you've used an arrow function above you could use one here too for consistency 🤷‍♂️

remybach: I would further argue that this can just be an anonymous function passed directly into the `map` 😬

PR #118, opened 2018-10-01
Remove the SSI message

liamkeaton:
This will never be used again


liamkeaton:
@jkerr321 Were there any assets on S3 that need to be removed?

jkerr321:
@liamkeaton no, as we didn't use any images in the end

PR #122, opened 2018-10-23
Don't import `./src/components/bottom/swg-entitlements-prompt/main` as it aint there   🐿 v2.10.2

maxcbc:
Don't import `./src/components/bottom/swg-entitlements-prompt/main.scss` as it aint there

 🐿 v2.10.2


liamkeaton:
It is there?

maxcbc:
but this is a scss import.

PR #129, opened 2018-11-15
Update top promo slot and add AB test logic

remybach:
🐿 v2.11.0

`control`:

`variant`:


alexnaish:
Don't know if its just the screenshot but I can see a teal bar under the image in the first screenshot. Could we make it 100% height to remove that?

remybach:
Ah, good spot! I had changed the way I was setting the height and forgot to go back and check that properly... will fix.

liamkeaton:
The second banner looks on the hoof and not centred

PR #146, opened 2019-04-16
Removes classes .variant-brexitCampaign and variant-techScrollAsia

CourtneyLO:
It appears that with the new version of n-gage the build is failing due to _verify_stylelint selector-class-pattern

failed-build

So I have made some changes to prevent that. Please check to see if the banners are correct

Brexit:

Screenshot 2019-04-16 at 07 55 41

Tech Scoll Asia:

Screenshot 2019-04-16 at 07 56 01

 🐿 v2.12.3


sbinlondon:
Thank you for this @CourtneyLO ! Apologies again for not doing the classes correctly. For future reference, where can I look to see what styles I should follow when adding to the repo? I based `techScrollAsia` off of the `Brexit` banner since it was one of the few I found that made significant style changes to the banner, but apparently the way the `Brexit` banner was built was not correct according to the linter?

PR #148, opened 2019-04-22
Tp/b2btrialmessages

tom-parker:


src/client/bottom-slot.js
```diff
@@ -3,7 +3,7 @@ const { generateMessageEvent, listen } = require('./utils');
 
 const BANNER_CLASS = 'n-messaging-banner';
 const BANNER_ACTION_SELECTOR = '[data-n-messaging-banner-action]';
-const BANNER_BUTTON_SELECTOR = `.${BANNER_CLASS}__button`;
+const BANNER_IMPERATIVE_ACTION_SELECTOR = `.${BANNER_CLASS}__action a`;
```
liamkeaton: Will this cause issues with tracking when it's not an `anchor` but a `button`, for instance https://github.com/Financial-Times/n-messaging-client/blob/master/templates/partials/bottom/onboarding-app.html#L19

tom-parker: I think that one might have been caught with the `BANNER_ACTION_SELECTOR` but this does seem a bit weird. Maybe I'm better off merging the button and link selections

kavanagh:
This PR was merged (indirectly via our Renovatebot PR) onto next-article without even a description or understandable commit message. Review and descriptions and reviews on this repo need to be better than this if we're going to work in this way.

tom-parker:
Apologies, descriptions and commit messages aside, do you think there's a better process for this?

kavanagh:
I think info on PRs will help a lot. I don't know a better process yet because I don't really know what we have (I'm slightly out of the loop). Let's discuss in email/person.

PR #155, opened 2019-05-15
Tracking code added to the Download links, message ID , appPromoMobile.

sentayhuJo:
🐿 v2.12.3
- Tracking code is added to the Download link for both `google play` and `Apple store`
- Google play download link corrected.


tom-parker:
I think we'll want these to be more specific to this journey. Using 'onsite-message' is a bit generic

sentayhuJo:
campaign tracker , `ct` changed from `onsite-message` to `onsite-app-promotion`

PR #162, opened 2019-07-03
New lazy messages added

sentayhuJo:
🐿 v2.12.3


sentayhuJo:
This message is built as part of the B2C registered user conversion initiative.

PR #180, opened 2019-09-06
New agenda campaign banner

CLTPayne:
### Description
Based on the current anonymous subscribe now banner, this is a new variant for the new agenda campaign release - https://trello.com/c/JcC0beJB/1462-1-new-agenda-campaign-release-plan. 

Normal State:
newAgendaBanner

Hover State:
New Agenda Hover State Banner

Notes:
* Very interested in feedback on how I've implemented the mixed font for the message. (New to handlebars so not sure what a best practise solution would be.)
* Discussed the hover state for the 'button' (it's a link) with Lucinda and agreed for it to be `lemon` and `black`.
* Still to do - media query or handling of smaller screen size. (I'm sorry!! I didn't manage to get this bit done so adding as a separate DOD to the ticket). 
* The 'button' has some javascript to extend the click area as per the current anonymous subscribe banner.
* As discussed with Umberto and Alex, have not yet added the variant to the `messageSlotTop` flag, leaving this to be done for the go-live.

[Trello Card](https://trello.com/c/sXV85Faf/1481-1-implement-campaign-banner)

## Tests:
Additional url now tested automatically by `pa11y` and passing.

### DIFF
--- a/src/components/top/new-agenda/main.scss
+++ b/src/components/top/new-agenda/main.scss
@@ -1,0 +1,46 @@
+.n-alert-banner--marketing__new-agenda {
+	display: none;
+
+	@include oGridRespondTo('M') {
+		display: block;
+		background-color: oColorsGetPaletteColor('lemon');
+	}
+
+	.n-alert-banner__content {
+		align-items: center;
+		display: flex;
+		height: 56px;
+		justify-content: center;
+		padding: 0 20px;
+	}
+
+	.n-alert-banner__cell {
+		padding: 0 15px;
+	}
+
+	.n-alert-banner__content-main {
+		@include oTypographyDisplay($scale: 4);
+		text-transform: uppercase;
+	}
+
+	.n-alert-banner__alternateFontStyle {
+		font-family: Metricweb;
+	}
+
+	.n-alert-banner__actions__primary {
+		@include oMessageButtonBase;
+		@include oButtonsTheme($theme: (background: 'white', accent: 'black', colorizer: 'primary'));
+		margin-right: 0;
+			// This is a hack around the fact that custom buttons do not
+			// support buttons with three colours. TODO: add this to o-buttons
+			// https://github.com/Financial-Times/o-buttons/issues/128
+		&:not([disabled]):hover {
+			background-color: oColorsGetPaletteColor('lemon');
+			color: oColorsGetPaletteColor('black');
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


remybach:
@CLTPayne: not that your way was necessarily wrong, but [I refactored the styling for the mixed font](https://github.com/Financial-Times/n-messaging-client/pull/180/commits/fd197eaf4028c1d800339389717e692df33602c7) so it makes more sense semantically (in my opinion). It's not necessarily neater from the styling perspective though.

I don't think there's a right way that works well on both fronts tbh, so I prefer the complexity to be on the code side rather than the markup. ¯\\\_(ツ)\_/¯

src/components/top/new-agenda/main.scss
```diff
@@ -0,0 +1,54 @@
...
+	.n-alert-banner__cell {
+		padding: 0 15px;
+	}
+
+	.n-alert-banner__content-main {
+		@include oTypographySans($scale: 3);
+		text-transform: uppercase;
+
+		em {
+			@include oTypographyDisplay($scale: 3);
```
alexnaish: Could this use the `oTypographySerif` mixin here?

src/components/top/new-agenda/main.scss
```diff
@@ -0,0 +1,54 @@
...
+	.n-alert-banner__content {
+		align-items: center;
+		display: flex;
+		height: 56px;
+		justify-content: center;
+		padding: 0 20px;
+	}
+
+	.n-alert-banner__cell {
+		padding: 0 15px;
```
alexnaish: If we've got `oSpacing`, it would be good to use it here otherwise its all good

src/components/top/new-agenda/main.scss
remybach: It could yeah... any particular reason why? It seems like a po-tay-to puh-ta-to situation to me.

alexnaish: Just as the mixin explicitly state the font and the style

remybach: Oddly enough, using `oTypographySerif($scale: 3);` makes it come out the wrong size: ¯\\\_(ツ)\_/¯

CLTPayne:
@remybach awesome! I didn't know about the emphasis semantic HTML element. I agree that makes the markup a lot cleaner.

PR #184, opened 2019-09-16
fixes myft disengaged tooltip spacing when followers > 0

maxcbc:
add missing space
instead of `read the latestbing, bang and bong stories` display `read the latest bing, bang and bong stories`


maxcbc:
> PR is good.
> 
> Side question: Can this code run with zero followed concepts? ie. is `content` = `Read the latest stories.` ever possible? Maybe we don't care, since ` ` will be rendered as ` `.

Nope it can't, it runs different code

PR #189, opened 2019-10-02
dont show myft header tooltip on refresh

maxcbc:
if document.referrer is blank (i.e refresh or bookmark), don't show the myft header tooltip

### DIFF
--- a/src/components/top/myft-disengaged-tooltip/main.js
+++ b/src/components/top/myft-disengaged-tooltip/main.js
@@ -5,7 +5,7 @@ import Cookies from 'js-cookie';
 
 const articleAddToMyftButton = document.querySelector('.topper__primary-theme .n-myft-follow-button');
 const headerMyFTLogo = document.querySelector('[data-trackable="my-ft"]');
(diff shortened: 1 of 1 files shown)


src/components/top/myft-disengaged-tooltip/main.js
```diff
@@ -5,7 +5,7 @@ import Cookies from 'js-cookie';
 
 const articleAddToMyftButton = document.querySelector('.topper__primary-theme .n-myft-follow-button');
 const headerMyFTLogo = document.querySelector('[data-trackable="my-ft"]');
-const externalReferer = !document.referrer || !(new URL(document.referrer).hostname.endsWith('ft.com'));
+const externalReferer = document.referrer && !(new URL(document.referrer).hostname.endsWith('ft.com'));
```
alexandercurtis:
```suggestion
const externalReferer = document.referrer && !(new URL(document.referrer).hostname.endsWith('.ft.com'));
```

otherwise hovercraft.com gets through

PR #194, opened 2019-10-31
Update ft weekend promo

saldixon:
new dynamic message using messaging guru.


tom-parker:
Looks good to me. Hopefully someone from CP can take a look too

PR #195, opened 2019-11-05
First pass 🐿 v2.12.5

saldixon:
Caters for a scenario not encountered before, where the CTA destination of an OSM is another page set up to display the same OSM.  In this case, the Spoor/Envoy pipeline does not react quickly enough to register a user action and therefore dismiss the message before the next page is rendered (and the message already re-displayed).

Storing OSM state and rules in the browser allows us to prevent this poor user experience.

This PR is a first pass to get us going quickly with ftWeekendPromo, and test this in production for a single message.  If successful, the intention is to extract the new functionality to be easily reusable by other messages.

NB.  A subsequent is also ready for view, and if you don't think it's too risky.. we could just go straight to this one instead, which is generic and will work for any message: 
 https://github.com/Financial-Times/n-messaging-client/pull/196/files

### DIFF
--- a/manifest.js
+++ b/manifest.js
@@ -99,7 +99,7 @@ module.exports = {
 		lazy: true
 	},
 	ftWeekendPromo: {
-		path: 'bottom/lazy',
+		path: 'bottom/ft-weekend-promo',
 		lazy: true
 	},
 	giftArticles: {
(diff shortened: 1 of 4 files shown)


maxcbc:
OSM?

saldixon:
> OSM?

Sorry.. 'On-site Message' :-)

PR #196, opened 2019-11-05
Local event counter

saldixon:
This supersedes the previous PR, which was cautiously trying to limit the potential blast radius to just ftWeekendPromo.
This PR allows n-messaging-client to configure event limits for each message and keep track of these locally, for a far quicker feedback loop and therefore better user experience.


src/client/utils.js
```diff
@@ -1,7 +1,20 @@
+const cookies = require('js-cookie');
+const _=require('lodash');
```
maxcbc: you should probably install and import just `lodash.set`and `lodash.get`here like a so:
```const set = require('lodash.set');```
This will reduce the bundle size

tom-parker:
LGTM

PR #207, opened 2019-11-26
Cascade

rowanmanning:
This is a monster, because we needed to completely remove custom classes. It might be easier for me to explain in-person, but I'll try to do it here.

- It turns out that `n-alert-banner` can be removed, which is great. This is because we're not using `o-message` classes directly
- I didn't want to make any overrides to `o-banner` or `o-message` classes, because this would be likely to negatively impact other parts of the site. So I created very specific classnames which reference the name of this repo. This at least means that if we have issues later it's super clear where the classes are defined (not the case before)
- Because the banner client-side JavaScript no longer accepts non-standard themes, there's some slightly dodgy wrangling of themes going on, and for the imperative banner we also add some classes after instantiation. I don't think there's a way around this stuff without rewriting most of `n-messaging-guru`

I've tested every banner variant locally, and they all look right with a few small spacing changes.


src/components/bottom/team-trial/main.scss
```diff
@@ -1,77 +1,58 @@
 @import 'o-overlay/main';
 @import 'o-loading/main';
 
+@include oOverlay();
+
+// Overlay override, I'm not sure why 😬
 .o-overlay {
-	@include oOverlay;
 	z-index: 1000;
```
notlee: This might effect other pages, think we can ad least stick it under `--n-overlay`?

README.md
```diff
@@ -167,7 +167,7 @@ If you need to turn a message off in production quickly, simply remove the varia
 ### The Solution
 
 - Message hierarchy as decided by "The Brain™" (`next-ammit-api`)
-- Consistent design and clear usage guidlines (`o-banner` and `n-alert-banner`)
+- Consistent design and clear usage guidlines (`o-banner` and `o-message`)
```
chee:
```suggestion
- Consistent design and clear usage guidelines (`o-banner` and `o-message`)
```

critical.scss
```diff
@@ -1,33 +1,38 @@
+$system-code: 'n-messaging-client' !default;
```
chee: is `n-messaging-client` a system? or would we want this to be set by the app?

rowanmanning: It's not, but we're doing this in components so that the demos can run and tests pass. We'd still want individual apps to override this value

edds: I got around this by introducing a new file in `n-content-body` so that the error will happen downstream. https://github.com/Financial-Times/n-content-body/commit/7c567b535f2b63c7c4d5bfe976bef798a53b658c

chee: i think Edd's solution to this is great so that app builders don't forget (or miss out on the chance to learn) to set their `$system-code`

rowanmanning: I agree, I didn't think of this. I've done the same now

src/components/bottom/cookie-consent/main.js
```diff
@@ -41,7 +41,7 @@ module.exports = function customSetup (banner, done) {
 		);
 
 		if (typeof CSS === 'undefined' || !CSS.supports('position', 'sticky')) {
-			bannerElem.classList.add('n-messaging-banner--fixed');
+			bannerElem.classList.add('o-banner--fixed');
```
notlee: so I know I already reviewed this but... `o-banner--fixed` isn't an o-banner class?

src/components/bottom/tech-scroll-asia/main.scss
```diff
@@ -22,15 +22,22 @@
...
-	.n-messaging-banner__content p {
+	.o-banner__content p {
 		padding-right: 40px;
 	}
 
-	.n-messaging-banner__button {
-		@include oButtonsTheme($theme: (background: (#3A70AF), accent: 'white', colorizer: 'primary'));
+	.o-banner__button {
+		@include oButtonsContent((
+			'type': 'primary',
```
notlee: so I know I already reviewed this but... I think we could reduce css output by setting `$include-base-styles: false`, (we wouldn't need to define the size then either, which overrides the default size)

src/components/bottom/cookie-consent/main.js
rowanmanning: It is now. Possibly I should keep it as a custom class though?
https://github.com/Financial-Times/n-messaging-client/pull/207/files#diff-a36753a1d7487efc8c4e8b0daedaac10L18-R18

notlee: Oh sorry, missed that, yeah I think that would be more clear

rowanmanning: Actually it's scoped inside `.cookie-banner`, so not adding a confusing class which would be available to others. Sorry just realised that link doesn't work, it's the file below which creates this class.

src/components/bottom/tech-scroll-asia/main.scss
rowanmanning: I'm trying to do this but I've spent _so_ long this morning getting demos set up on my home mac 😩

rowanmanning: Done!

PR #209, opened 2019-11-27
new message on message slot top added for nbe auto subscription journ…

sentayhuJo:
…ey 🐿 v2.12.5
Screenshot 2019-11-27 at 09 09 15

### DIFF
--- a/src/components/top/nbe-auto-sub/main.js
+++ b/src/components/top/nbe-auto-sub/main.js
@@ -1,0 +1,17 @@
+module.exports = function customSetup (banner, done) {
+
+	const bannerActions = banner.messageElement.querySelector('.n-alert-banner__actions');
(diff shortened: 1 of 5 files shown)


src/components/top/nbe-auto-sub/main.js
```diff
@@ -0,0 +1,17 @@
...
+		// Hit test to see if the click originated on expanded click area
+		if (clickEvent.target !== bannerButton) {
+			clickEvent.stopPropagation();
+			bannerButton.click();
+			return false;
+		}
+	});
+
+	done();
+};
```
sjparkinson: It's not clear to me what the purpose of this code is. It would be helpful to include a comment explaining what it is intended to do.

sentayhuJo: It is not used on this particular message implementation, Only the `main.scss` is use for styling the banner. It should be removed.

PR #212, opened 2019-12-03
New messages: fastFt and marketsData

saldixon:
These are two brand new static messages for the bottom slot (frontpage,stream pages, and article pages).  

They have been built for a new Envoy journey (nba-test-1.1vs2) whose purpose is to evaluate two Datascience models, both of which predict the 'next best action' for a particular user to maximise their engagement, but using different algorithms. Envoy displays an onsite message encouraging the 'next best action' (NBA) for a group of **~60k pre-selected candidates for the test**. The clickthrough rates and subsequent RFV scores are then used to determine if one model predicted better than the other.  More info on the initiative here:  https://docs.google.com/presentation/d/1Cd9pvhmrnYZLhV0v9I9YmYXLno0l_hb7mfNkUk_e2dE/edit#slide=id.g50dea84098_0_0

Screenshot of the new messages:

image
image
image
image


src/components/bottom/fast-ft/main.scss
```diff
@@ -0,0 +1,60 @@
...
+		color: black;
+		background: white;
+
+		@include oGridRespondTo($from: M) {
+			background-image: url("https://www.ft.com/__origami/service/image/v2/images/raw/https%3A%2F%2Fenvoy-fast-ft.s3-eu-west-1.amazonaws.com%2FfastFT_promo_message_graphic.png?source=ip-envoy");
+			background-size: contain;
+			background-repeat: no-repeat;
+			background-position-x: calc(100% + 1px); // a workaround to remove 1px gap on the right/bottom
+			background-position-y: calc(100% + 1px);
+			background-size: 45%;
```
umbobabo: In term of performance and quality, I would suggest that you use images already resized if possible and use them at 100% of their size. Downscaling can produce undesired effect in term of quality. You can [resize the image via the origami service](https://www.ft.com/__origami/service/image/v2/#query-parameters). This would reduce the amount of data you send to the browser. You should consider that you are serving a png with transparency that does seem optimised (192 kb is a lot for your purpose). If the background is predictable would be better using an image without transparency cause it will save a lot of size. Anyway, just reducing the image and use the background at 100% should improve performances.

src/components/bottom/fast-ft/main.scss
```diff
@@ -0,0 +1,60 @@
...
+		}
+	}
+
+	.o-banner__heading h1 {
+		font-family: FinancierDisplayWeb, serif;
+		font-size: 28px;
```
umbobabo: You should probably use [o-typography sizes](https://registry.origami.ft.com/components/o-typography@6.0.1) because it enforces constistency

src/components/bottom/markets-data/main.scss
```diff
@@ -0,0 +1,65 @@
...
+		color: black;
+		background: white;
+
+		@include oGridRespondTo($from: M) {
+			background-image: url("https://www.ft.com/__origami/service/image/v2/images/raw/https%3A%2F%2Fenvoy-markets-data.s3-eu-west-1.amazonaws.com%2Fmarkets_data.png?source=ip-envoy");
+			background-size: contain;
+			background-repeat: no-repeat;
+			background-position-x: calc(100% + 1px); // a workaround to remove 1px gap on the right/bottom
+			background-position-y: calc(100% + 1px);
+			background-size: 40%;
```
umbobabo: Same as per other banner

src/components/bottom/fast-ft/main.scss
```diff
@@ -0,0 +1,60 @@
...
+			background-color: white;
+			border-color: black;
+		}
+	}
+
+
+
+
+
+
```
umbobabo: A lot of unnecessary spaces

src/components/bottom/markets-data/main.scss
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	.o-banner__heading h1 {
+		font-family: FinancierDisplayWeb, serif;
+		font-size: 28px;
+		line-height: 28px;
+	}
+
+	.o-banner__product_name {
+		color: #29C697;
```
umbobabo: I think we should use colors from o-color https://registry.origami.ft.com/components/o-colors@5.0.2

src/components/bottom/markets-data/main.scss
```diff
@@ -0,0 +1,65 @@
...
+	.o-banner__heading:after {
+		margin-top: 8px;
+		margin-bottom: 16px;
+		content: '';
+		display: block;
+		width: 100px;
+		border-bottom: 0;
+	}
+
+
```
umbobabo: 1 unnecessary space

src/components/bottom/markets-data/main.scss
```diff
@@ -0,0 +1,65 @@
...
+			background-color: white;
+			border-color: black;
+		}
+	}
+
+
+
+
+
+
```
umbobabo: Too many spaces.

PR #222, opened 2019-12-06
The onsite message(appPromoMobile) styling issue on mobile device is …

sentayhuJo:
…fixed 🐿 v2.12.5
This is to fix the styling issue of appPromoMobile message on the mobile device.

### It was look like this

Old

### Now look like this
new

### DIFF
--- a/src/components/bottom/app-promo-mobile/main.scss
+++ b/src/components/bottom/app-promo-mobile/main.scss
@@ -1,16 +1,18 @@
 .n-messaging-client-messaging-banner--app-promo-mobile {
 	.app-promo-mobile__device {
-		left: 70px;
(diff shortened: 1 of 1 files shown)


src/components/bottom/app-promo-mobile/main.scss
```diff
@@ -1,16 +1,18 @@
...
 	.app-promo-mobile__device {
-		left: 70px;
+		left: 90px;
 		top: 20px;
 		position: absolute;
 	}
-	.o-banner__action {
+
+	.o-banner__content--img {
+		margin-left: 238px;
```
chee: should we have a comment here that explains this number?

PR #226, opened 2019-12-11
Remove lodash.

notlee:
It's a major version behind and the Snyk report for dependent projects
is fretting about possible vulnerabilities. It's only used to get and
set properties in one place so let's not use lodash here.

### DIFF
--- a/src/client/utils.js
+++ b/src/client/utils.js
@@ -1,6 +1,4 @@
 const cookies = require('js-cookie');
-const set = require('lodash.set');
-const get = require('lodash.get');
 
 const manifest = require('../../manifest');
 const LOCAL_COUNTER_COOKIE_NAME = 'nMessagingEventCounter';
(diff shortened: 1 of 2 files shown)


edds:
Think there is one more usage of `get` that needs updating. But 🎉  this is great! THANK YOU!

src/client/utils.js
```diff
@@ -10,14 +8,12 @@ const dispatchEvent = (event) => {
...
-const getMessageRules = (messageId) => {
-	return get(manifest, `${messageId}.eventRules`);
-};
+const getMessageRules = messageId => manifest[`${messageId}.eventRules`];
 
 const updateLocalCounter = (messageId, event) => {
 	const currentCounts = cookies.get(LOCAL_COUNTER_COOKIE_NAME) ? JSON.parse(cookies.get(LOCAL_COUNTER_COOKIE_NAME)) : {};
-	const currentCount = get(currentCounts, `${messageId}.${event}`) || 0;
-	set(currentCounts, `${messageId}.${event}`, currentCount + 1);
+	const currentCount = currentCounts[`${messageId}.${event}`] || 0;
```
JakeChampion: Bracket notation does not support deep property gets such as this. We will need to rewrite this to be:
```suggestion
	const currentCount = currentCounts[messageId] != null && currentCounts[messageId][event] != null ?  currentCounts[messageId][event] :  0;
```

notlee: thanks, I only just noticed this. Just going to check what events this is dealing with, in case they have a `.` in the event name which is important

notlee: they do not!

src/client/utils.js
```diff
@@ -10,14 +8,12 @@ const dispatchEvent = (event) => {
...
+	currentCounts[`${messageId}.${event}`] = currentCount + 1;
```
JakeChampion:
```suggestion
	currentCounts[messageId] = currentCounts[messageId] || {};
	currentCounts[messageId][event] = currentCounts[messageId][event] || 0;
	currentCounts[messageId][event] = currentCount + 1;
```

notlee: Thanks, I totally missed the need for this 😅

notlee:
Sorry @edds this is not what I thought it was -- loadash is replaced properly now

PR #235, opened 2020-01-14
Missing formAction added to Imperative option.

sentayhuJo:
Missing imperative formAction added in order to render form html tag on onsite message 🐿 v2.12.5


sentayhuJo:
> Looks good but should `formEncoding` and `formMethod` be added also?

Yes, you are right , i have to include them as well. Thank you

PR #237, opened 2020-01-16
Adds .default to the o-message require statement

jenniferemshepherd:
to resolve ES6/CJM errors in Page Kit

### DIFF
--- a/src/client/top-slot.js
+++ b/src/client/top-slot.js
@@ -1,4 +1,5 @@
-const nAlertBanner = require('o-message').default;
+let nAlertBanner = require('o-message');
+nAlertBanner = nAlertBanner.default || nAlertBanner;
 const { generateMessageEvent, listen, messageEventLimitsBreached } = require('./utils');
 
 const ALERT_BANNER_CLASS = 'o-message';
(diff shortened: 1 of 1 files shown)


src/client/top-slot.js
```diff
@@ -1,4 +1,4 @@
-const nAlertBanner = require('o-message');
+const nAlertBanner = require('o-message').default;
```
pixelandpage: It seems confusing to me that this variable is called `nAlertBanner` when it comes from o-message, rather than the [n-alert-banner](https://github.com/Financial-Times/n-alert-banner) repo. Could this to be renamed `Message` to match the name of the default export in o-message (https://github.com/Financial-Times/o-message/blob/master/main.js#L11) and avoid some potential confusion?

jenniferemshepherd: sure - it's not my repo so I'm keen not to make changes to someone else's code base where they might have reasons to do it this way.  I'm happy to make the change though if you think it's better.

pixelandpage: That's a great point! 

As we're not exporting this value it should be safe to change, but you could ask for a review from someone in the team too to make sure.

remybach: I'm happy for this to be renamed 👍

PR #242, opened 2020-01-20
One click daily digest

bjfletcher:
Replace action link with an "in-page" button action, to sign up to the daily digest. The myFT code for the sign up is untested.


src/components/bottom/daily-digest/main.js
```diff
@@ -0,0 +1,32 @@
...
+		function handleSignUpClick (evt) {
+			evt.preventDefault();
+			const conceptId = document.documentElement.dataset.conceptId;
+			if (conceptId) {
+				function addUserToConcept () {
+					return myft.add('user', null, 'followed', 'concept', conceptId);
+				}
+				function logError (err) {
+					/* eslint no-console:0 */
+					console.log({ info: 'could not add user to cocnept', conceptId, err });
```
CourtneyLO: I believe there is a typo here:
```suggestion
					console.log({ info: 'could not add user to concept', conceptId, err });
```

bjfletcher:
@sentayhuJo I would do the following:

1. instead of generating HTML in JS, do it like this:

```
<div class="....front-banner ...--is-visible">...</div>
<div class="...back-banner">...</div>
```

Then when the subscription is successful, change `--is-visible` from `front-banner` to `back-banner` using JS & `classList`.

In this way, you can write all HTML in HTML, not HTML in JS. You can also do nice CSS transitions like flipping, should Emma want it!

2. instead of custom styling, always use Origami's styling:

https://registry.origami.ft.com/components/o-buttons@6.0.5
https://registry.origami.ft.com/components/o-icons@6.0.0 (see the "tick" one)
https://registry.origami.ft.com/components/o-colors@5.0.5 (if you're unsure which ones, ask Emma)

In this way, the styling, the look and feel will be consistent with the rest of FT.

bjfletcher:
PS I've removed unused modules (`get-token` and `sign-up`)

secret-squirrel.js
```diff
@@ -16,7 +16,8 @@ module.exports = {
 			'eb62aef2-4bdc-0f31-a2c5-6532b8e17896', // templates/partials/bottom/content-message.html:15
 			'5f60b8b4-cbf0-18d7-df41-9caa1171e8c1', // templates/partials/top/anon-subscribe-now.html:5|18
 			'a44c9005-2a9c-fe2a-9140-1311ff87f25f', // templates/partials/top/print-banner-usa.html:4
-			'79c8c0c5-1686-7f28-3aeb-c14688d59749' // templates/partials/top/uk-election.html:4
+			'79c8c0c5-1686-7f28-3aeb-c14688d59749', // templates/partials/top/uk-election.html:4
+			'82645c31-4426-4ef5-99c9-9df6e0940c00' // src/components/bottom/daily-digest/main.js: 9
```
tom-parker: I'm not sure this needs to be here

src/components/bottom/daily-digest/main.js
```diff
@@ -0,0 +1,56 @@
+import myft from 'next-myft-client';
+import Cookies from 'js-cookie';
+
+module.exports = function customSetup (banner, done) {
+	if (window.FT && window.FT.flags && window.FT.flags.oneClickDailyDigest) {
```
tom-parker: Do we definitely want to scope this message to a single flag? This custom JS setup should only be triggered if a user should be seeing this message so I don't think we need this double check?

secret-squirrel.js
sentayhuJo: This offer Id is added by @remybach , https://github.com/Financial-Times/n-messaging-client/commit/03e830d16ebaecf3f10b927a1b60490a1f26203b

src/components/bottom/daily-digest/main.js
saldixon: wasn't this purely for testing purposes?  this flag?  to be removed once we had confidence in the one-click behaviour?

secret-squirrel.js
tom-parker: Yep, not the '82645c31-4426-4ef5-99c9-9df6e0940c00' value though. That's only appeared in this PR but I don't think it's actually used anywhere else in the code so we can probably remove it

src/components/bottom/daily-digest/main.js
```diff
@@ -0,0 +1,58 @@
+import myft from 'next-myft-client';
```
tom-parker: I don't think this file is needed for the `oneclick-daily-digest` messageID

src/components/bottom/daily-digest/main.scss
```diff
@@ -0,0 +1,58 @@
+.o-banner_success-icon-tick {
```
tom-parker: I don't think this file is needed for the `oneclick-daily-digest` messageID

templates/partials/bottom/daily-digest.html
```diff
@@ -1,17 +1,31 @@
 <div class="n-messaging-client-messaging-banner">
```
tom-parker: There shouldn't be any changes to this file if we're using the new `oneclick-daily-digest` message ID

src/components/bottom/daily-digest/main.js
sentayhuJo: did you mean for `daily-digest` message ID, coz originally we don't have such file for the daily digest.

src/components/bottom/daily-digest/main.scss
sentayhuJo: dito

sentayhuJo:
This PR is close because we need to tidy up the change.

PR #243, opened 2020-01-30
Adds ability to track dynamic content in OSMs

tom-parker:
When a view event is fired from n-messaging-client we have no way of knowing the dynamic content that the message contained. This is ok when the messages are static and the contents remains the same but doesn't work for dynamic messages with changing content.

As an example, if we're pushing topic recommendations into a message, these could be different for each user. We need to be able to see which topics were present when a user views the message.

For now, we're making use of the `messaging_variant` field that is already present in the tracking events and used by the analytics team. By default, this field contains the message name -
image

This change allows `messaging_variant` to contain a string that explains the dynamic content contained in the message. e.g
image

🐿 v2.12.5

### DIFF
--- a/src/client/bottom-slot.js
+++ b/src/client/bottom-slot.js
@@ -13,7 +13,8 @@ const BOTTOM_SLOT_FLAG = 'messageSlotBottom';
 
 module.exports = function ({ config={}, guruResult, customSetup }={}) {
 	let banner;
-	const trackEventAction = config.name && generateMessageEvent({ messageId: config.name, position: config.slot, variant: config.name, flag: BOTTOM_SLOT_FLAG });
+	const variant = (guruResult && guruResult.renderData && guruResult.renderData.dynamicTrackingData) || config.name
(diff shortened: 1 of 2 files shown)


src/client/bottom-slot.js
```diff
@@ -13,7 +13,8 @@ const BOTTOM_SLOT_FLAG = 'messageSlotBottom';
 
 module.exports = function ({ config={}, guruResult, customSetup }={}) {
 	let banner;
-	const trackEventAction = config.name && generateMessageEvent({ messageId: config.name, position: config.slot, variant: config.name, flag: BOTTOM_SLOT_FLAG });
+	const variant = (guruResult && guruResult.renderData && guruResult.renderData.dynamicTrackingData) || config.name
```
umbobabo:
```suggestion
	const variant = (guruResult && guruResult.renderData && guruResult.renderData.dynamicTrackingData) || config.name;
```

tom-parker: 👌

PR #245, opened 2020-02-06
New message added for oneclick daily email digest sign up 🐿 v2.12.5

sentayhuJo:
This new message is added for one click daily email digest subscription. it is intended for Envoy journey. 

Screenshot 2020-02-06 at 15 54 21

Once the user click the sign up button, it automatically subscribed to the daily email digest newsletter. And try the user to be more engage to the newsletters on myFt.

Screenshot 2020-02-06 at 15 55 50


tom-parker:
LGTM

src/components/bottom/oneclick-daily-digest/main.scss
```diff
@@ -0,0 +1,58 @@
...
+	float: left;
+}
+
+.o-banner_button-browse {
+	@include oButtonsContent($opts: ('type': 'secondary'));
+	color: oColorsMix('black', 'jade', $percentage: 43);
+	border-color: oColorsMix('black', 'jade', $percentage: 43);
+	font-family: MetricWeb,sans-serif;
+	-webkit-font-smoothing: antialiased;
+	font-size: 16px;
```
CourtneyLO: Can you use o-typography here https://github.com/Financial-Times/o-typography?

src/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,56 @@
...
+				try {
+					return myft.add('user', null, 'preferred', 'preference', 'email-digest', {
+						token: getCSRFToken,
+						_rel:{
+							type: 'daily',
+							sendTime:'every morning'
+						}
+					});
+				} catch(e) {
+					return e;
```
alexnaish: If this was an async function - if you return `e` here, the promise chain assumes that the error has been handled and will never go into your catch block - either log and re-throw the error or don't catch it and let it bubble up.

On a side note - since this isn't marked as an `async` function, what here would through a synchronous error?

src/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,56 @@
...
+			evt.preventDefault();
+
+			function getCSRFToken () {
+				const desiredTokenLength = 36;
+				const token = Cookies.get('FTSession_s') || Cookies.get('FTSession');
+				const csrfToken = token ? token.slice(-desiredTokenLength) : '';
+				return csrfToken;
+			}
+
+			function addUserToDigest () {
```
alexnaish: These functions should ideally not be defined in the click handler itself, makes it harder to read and maintain in my opinion. If they were pulled out into their own file it'd make it easier to understand what is responsible for what and it would be possible to test this logic

src/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,56 @@
...
+
+module.exports = function customSetup (banner, done) {
+	if (window.FT && window.FT.flags && window.FT.flags.oneClickDailyDigest) {
+		function loadSuccessContent () {
+			const outerContainer = banner.bannerElement.querySelector('.o-banner__outer');
+			outerContainer.classList.add('o-banner_success-background');
+			const closeButton = banner.bannerElement.querySelector('.o-banner__close');
+			closeButton.classList.add('o-banner__close-color');
+			const signupContent = banner.bannerElement.querySelector('.o-banner_signup-content');
+			signupContent.classList.add('--is-visible');
```
alexnaish: Not sure what this class is - its not a modifier (http://getbem.com/naming/) and it adds `display: none` which is the opposite of being visible. Could we update this to be inline with both BEM and functionality please?

templates/partials/bottom/oneclick-daily-digest.html
```diff
@@ -0,0 +1,31 @@
...
+	}}
+	<div class="o-banner__inner" data-o-banner-inner="">
+		<div class="o-banner_signup-content">
+			<div class="o-banner__content">
+				<p><strong>Sign up to the myFT daily digest</strong></p>
+				<p>Receive the latest news from topics you follow on myFT straight to your inbox</p>
+			</div>
+			<div class="o-banner__actions">
+				<div class="o-banner__action">
+					<button class="o-banner_button-signup" data-n-messaging-action> Sign up </button>
```
alexnaish: This here doesn't specify a button time and doesn't work in core - the message is still rendered with Javascript disabled but doesn't actually do anything. This should maybe be wrapped in a form so that even without JS it will still function as expected.

src/components/bottom/oneclick-daily-digest/main.scss
sentayhuJo: Great ! i can do that

src/components/bottom/oneclick-daily-digest/signUp-to-digest.js
```diff
@@ -0,0 +1,22 @@
+import myft from 'next-myft-client';
+import Cookies from 'js-cookie';
+
+module.exports = function signUpToDigest () {
```
alexnaish: Can we use `export default` instead of `module.export` here (and in other modified files using import/export) please - makes it more consistent.

templates/partials/bottom/oneclick-daily-digest.html
alexnaish: This comment still stands - the message is rendered without JS but is reliant on JS in order to function. There needs to be some thought about how it should fallback please.

src/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = function customSetup (banner, done) {
+	if (window.FT && window.FT.flags && window.FT.flags.oneClickDailyDigest) {
+		function loadSuccessContent () {
+			const outerContainer = banner.bannerElement.querySelector('.o-banner__outer');
+			outerContainer.classList.add('o-banner_success-background');
+			const closeButton = banner.bannerElement.querySelector('.o-banner__close');
+			closeButton.classList.add('o-banner__close-color');
+			const signupContent = banner.bannerElement.querySelector('.o-banner_signup-content');
+			signupContent.classList.add('o-banner_is-visible');
```
alexnaish: Going to be picky here (sorry).

The "is visible" state is a modification of the block element - with BEM this means the class should be structured `.o-banner--is-visible` with the modifier stated via the `--` syntax.

However, the presence of the class implies to me that the block should be visible is actually false - as the class adds `display: none` - the class name should be updated to something akin to `.o-banner--is-hidden` as that accurately describes what it does.

src/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,37 @@
...
+import signUpToDigest from './signUp-to-digest';
+
+module.exports = function customSetup (banner, done) {
+	if (window.FT && window.FT.flags && window.FT.flags.oneClickDailyDigest) {
+		function loadSuccessContent () {
+			const outerContainer = banner.bannerElement.querySelector('.o-banner__outer');
+			outerContainer.classList.add('o-banner_success-background');
+			const closeButton = banner.bannerElement.querySelector('.o-banner__close');
+			closeButton.classList.add('o-banner__close-color');
+			const signupContent = banner.bannerElement.querySelector('.o-banner_signup-content');
```
alexnaish: Minor stylistic preference - could we declare all the `const` at the top of the function and then perform the actions on them?

ie:

```
const outerContainer = banner.bannerElement.querySelector('.o-banner__outer');
const closeButton = banner.bannerElement.querySelector('.o-banner__close');
const signupContent = banner.bannerElement.querySelector('.o-banner_signup-content');
// ... class modifications go below
```

In an ideal world, the function itself should be pulled outside of the `customSetup` function and have the bannerElement passed in as a parameter - this would mean your `customSetup` is lean and clean ;)

src/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,37 @@
+// import myft from 'next-myft-client';
+// import Cookies from 'js-cookie';
```
alexnaish: No need for these comments, they don't give us anything - I'd remove them please.

src/components/bottom/oneclick-daily-digest/main.scss
```diff
@@ -0,0 +1,57 @@
+.o-banner_success-icon-tick {
+	@include oIconsContent(
+		$icon-name: 'tick',
+		$color: oColorsMix('black', 'jade', $percentage: 43),
```
alexnaish: Why so many magic numbers in here? Why 43 and 27? Are there not existing origami mixin's that allow us to be more inline with defaults?

umbobabo:
@sentayhuJo Can this one be closed?

sentayhuJo:
Sure ! , @umbobabo , Thank you

PR #254, opened 2020-02-21
Daily digest one-click banner

sentayhuJo:


client/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,33 @@
+import signup from './signup';
+export default async (banner, done) => {
```
bjfletcher: sorry to be a pain, but async/await isn't something we can use for client-side JS, unfortunately... so have to use promise/then/catch instead

client/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,33 @@
...
+	//if (window.FT && window.FT.flags && window.FT.flags.oneClickDailyDigest) {
+	if(true) {
+		function loadSuccessContent () {
+			const outerContainer = banner.bannerElement.querySelector('.o-banner__outer');
+			outerContainer.classList.add('o-banner_success-background');
+			const closeButton = banner.bannerElement.querySelector('.o-banner__close');
+			closeButton.classList.add('o-banner__close-color');
+			const signupContent = banner.bannerElement.querySelector('.o-banner_signup-content');
```
bjfletcher: convention is double underline __ and not single underline _ in between parent and child components with the class names

client/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,33 @@
...
+			signupContent.classList.add('o-banner_is-visible');
+			const successContent = banner.bannerElement.querySelector('.o-banner_success-content');
+			successContent.classList.remove('o-banner_is-visible');
+		}
+
+		async function handleSignUpClick (evt) {
+			evt.preventDefault();
+			function logError (err) {
+				/* eslint no-console:0 */
+				console.log({ info: 'could not add user to the daily digest', err });
```
bjfletcher: i wonder if this should be console.warn instead of console.log

client/components/bottom/oneclick-daily-digest/main.js
```diff
@@ -0,0 +1,33 @@
...
+			}
+
+			await signup().then(loadSuccessContent).catch(logError);
```
bjfletcher: as mentioned earlier, async/await can't be used client-side

client/components/bottom/oneclick-daily-digest/main.scss
```diff
@@ -0,0 +1,57 @@
+.o-banner_success-icon-tick {
```
bjfletcher: i wonder if you could use Rowan's new theme generator instead of writing the styles here? example: https://github.com/Financial-Times/n-messaging-client/blob/one-click-myft-follows/client/themes.scss

client/components/bottom/oneclick-daily-digest/main.scss
```diff
@@ -0,0 +1,57 @@
...
+	border-color: oColorsMix('black', 'jade', $percentage: 43);
+}
+
+.o-banner_success-background {
+	background: oColorsMix('jade', $percentage: 20);
+	color: oColorsMix('black', 'jade', $percentage: 43);
+}
+
+.o-banner_signup-content {
+	position: relative;
```
bjfletcher: are these next two styles needed? could o-banner--small for example be used instead?

client/components/bottom/oneclick-daily-digest/signup.js
```diff
@@ -0,0 +1,15 @@
+import myftClient from 'next-myft-client/myft-bower';
+
+const addUserToDigest = async () => {
+	return await myftClient.add('user', null, 'preferred', 'preference', 'email-digest', {
```
bjfletcher: is just enabling daily digest the aim here? is the user already following the topic? i.e. we don't want to use the `followPlusDigestEmail` API?

demos/webpack.config.js
```diff
@@ -1,9 +1,16 @@
+const xEngine = require('@financial-times/x-engine/src/webpack');
+
 module.exports = {
 	mode: 'development',
 	resolve: {
 		modules: [__dirname + '/../bower_components', __dirname + '/../node_modules'],
-		mainFiles: ['index', 'main']
+		descriptionFiles: ['bower.json', 'package.json'],
```
bjfletcher: hopefully these are on `master` by now and you can merge to latest

package.json
```diff
@@ -11,17 +11,25 @@
     "prepare": "npx snyk protect || npx snyk protect -d || true"
   },
   "license": "MIT",
+  "x-dash": {
```
bjfletcher: it looks like you've branched off my myFT follows branch? can you rebase to master instead?

client/components/bottom/oneclick-daily-digest/signup.js
sentayhuJo: The user doesn't follow any topic at this point, yes i don't think we need to use `followPlusDigestEmail` API

demos/webpack.config.js
sentayhuJo: Yes, i have seen it. will pick up the change

PR #256, opened 2020-02-25
Removes the onboarding app promo message

tom-parker:
The Envoy journey that made use of this has been modified to use another
app promotional message so this message is no longer needed.

In theory, this should mean that the [email-app-links](https://github.com/Financial-Times/next-email-app-links) application can now be retired.

🐿 v2.12.5


tom-parker:
From the bit of digging I did, it seemed that all the references to `/email-app-links` were from archived repos so I think it's safe but worth double checking.

PR #265, opened 2020-03-11
A new coronavirus newsletter promotion

bjfletcher:
This is identical to the existing Tech Scroll Asia promotion message, with the following changes:

- copy (text)
- newsletter ID in link (coronovirus's instead of Tech Scroll Asia's)
- styling (use slate/lemon colours instead of Tech Scroll Asia's
custom styling)

Desktop:
image

Mobile:
image

### DIFF
--- a/client/themes.scss
+++ b/client/themes.scss
@@ -10,7 +10,7 @@
 	text-color: oColorsMix('black', 'jade', $percentage: 43)
 ));
 
-@include oBannerAddTheme('pikachu', (
+@include oBannerAddTheme('slate-lemon', (
 	background-color: oColorsByName('slate'),
 	text-color: oColorsByName('white'),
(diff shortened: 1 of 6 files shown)


client/themes.scss
```diff
@@ -9,3 +9,10 @@
 	background-color: oColorsMix('jade', $percentage: 20),
 	text-color: oColorsMix('black', 'jade', $percentage: 43)
 ));
+
+@include oBannerAddTheme('pikachu', (
```
umbobabo: As mentioned in our chat I do believe a meaningful name would be better. My suggestion is something like `lemonslate` cause it's a very custom template you are creating here.

bjfletcher: changed to `slate-lemon`

PR #266, opened 2020-03-12
Support a new Envoy "Next Best Action" journey

bjfletcher:
The technical approach for delivering 7 different messages based on "next best action" rules in Envoy is as follows:

- server-side render a "placeholder" message, that is invisible, to reserve space on the page, to prevent "jerky" page loading behaviour – i.e., we want everything on the page to stay where they are!
- client-side fetch (from the guru) the desired message and render it into the placeholder message
- flip between 3 different sub-messages: 1. click-to-action (initial view), 2. success (shades of green), and 3. error (shades of red) if click-to-action fails

image


bjfletcher:
Using the smallest message copy, which is fastFT. Here are the screenshots...

Desktop:
image

iPad:
image

small iPhone:
image

large iPhone:
image

(note that with the large iPhone, there's one blank line between copy and CTA button, because the placeholder copy takes up 3 lines whereas the actual copy takes up 2 lines)

bjfletcher:
I have tested this on `next-article` using `next-router` and `next-messaging-guru`, testing both some existing and this new message with the top slot, and some existing message with the bottom slot. All look fine & no errors in the JS console. I've made a video (100mb!) demoing the different NBA messages if you'd like to see a copy, let me or @saldixon know.

client/components/top/next-best-action/actions.js
```diff
@@ -1,8 +1,39 @@
+import { client as myft } from 'n-myft-ui/myft';
+import getCsrfToken from 'n-myft-ui/myft/ui/lib/get-csrf-token';
 import dailyDigestSignUp from '../../../lib/daily-digest';
 
 export function newsletter (guruResult) {
-	const url = guruResult.renderData.ctaActionHref;
-	return fetch(url)
+	function getUserId () {
+		// if there's a better way than using myFT, I'd like to know!
```
saldixon: I can easily get the guru to return it in the payload?

client/components/top/next-best-action/actions.js
```diff
@@ -1,8 +1,39 @@
...
+		return myft.init()
+			.then(function withMyFt () {
+				return myft.userId;
+			});
+	}
+	function getNewsletterId (guruResult) {
```
saldixon: no need for this.  The guru already returns the newsletter id, rather than a subscribe url.  Sample renderData:

    "renderData": {
        "reservedSpace": true,
        "nbaMessageId": "newsletter",
        "dynamicTrackingData": "nbaNewsletter,56d42625a2b6c30300fd5748",
        "newsletterName": "FT Weekend",
        "newsletterId": "56d42625a2b6c30300fd5748"
    }

client/components/top/next-best-action/actions.js
bjfletcher: @saldixon or guru could get us the myFT URL? would simplify a lot of this

bjfletcher:
Thanks @alexnaish. I've added code comments to that actions.js file you mentioned, and also the other JS files. See last commit.

PR #267, opened 2020-03-13
Add KPMG message

remybach:
[Ticket](https://financialtimes.atlassian.net/browse/AC-256)

Related [`next-ammit-api` PR](https://github.com/Financial-Times/next-ammit-api/pull/646)

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -2,7 +2,6 @@ $system-code: 'n-messaging-client';
 
 @import './client/themes';
 @import './client/components/top/anon-subscribe-now/main';
-@import './client/components/bottom/team-trial/main';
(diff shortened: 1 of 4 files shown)


main.scss
```diff
@@ -2,7 +2,6 @@ $system-code: 'n-messaging-client';
 
 @import './client/themes';
 @import './client/components/top/anon-subscribe-now/main';
-@import './client/components/bottom/team-trial/main';
```
umbobabo: Good catch 👍

remybach: It was hard to miss to be honest... the demo app wasn't building with this in there 😅

PR #278, opened 2020-04-24
ACQ-101: Add custom tracking context for marketingPopupPrompt, anonSubscribeNow, and printBannerUsa

remybach:
[Ticket](https://financialtimes.atlassian.net/browse/ACQ-101)

### DIFF
--- a/client/utils.js
+++ b/client/utils.js
@@ -37,16 +37,16 @@ const updateMessageEventCount = (messageId, event) => {
 };
 
 module.exports = {
-	generateMessageEvent: function ({ messageId, position, flag, variant }={}) {
+	generateMessageEvent: function ({ messageId, position, flag, trackingContext, variant }={}) {
 		return function (action, trackingAttr) {
-			const detail = {
+			const detail = Object.assign({
 				category: 'n-messaging',
 				action: action,
 				messaging: messageId,
 				messaging_position: position,
 				messaging_flag: flag,
 				messaging_variant: variant
-			};
+			}, trackingContext);
 			if (trackingAttr)
 				detail.messaging_action = trackingAttr;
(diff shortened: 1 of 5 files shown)


client/utils.js
```diff
@@ -37,16 +37,16 @@ const updateMessageEventCount = (messageId, event) => {
 };
 
 module.exports = {
-	generateMessageEvent: function ({ messageId, position, flag, variant }={}) {
+	generateMessageEvent: function ({ messageId, position, flag, trackingContext, variant }={}) {
```
umbobabo, line 40: I would set a default value for trackingContext. 
```suggestion
	generateMessageEvent: function ({ messageId, position, flag, trackingContext = {}, variant }={}) {
```
or you could have an error when you do object.assign if that is not passed for other messages

remybach: > `Object.assign()` does not throw on null or undefined sources.
> - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

I did test this on other messages and they worked fine. Can't hurt to put a default in though so I'll update it.

umbobabo: Don't change it then. I thought it was throwing on `undefined`

umbobabo:
Approved @remybach, pls do not release now but do it on Monday (except in case of extreme emergency, top priority)

remybach:
@umbobabo You must think me a mad man to release this late on a Friday 🤣

umbobabo:
> @umbobabo You must think me a mad man to release this late on a Friday 🤣

I think of you as a wise man instead, but I also know that sometimes we got super-duper-urgent-top-priority-we-cannot-wait-till-monday requests around messaging. 😉

ludrob:
8 km v thymus hi +hh+y 6u.                                    
   N.      6 bbb 5

PR #279, opened 2020-04-28
Make myFT disengaged tooltip as a permanent feature

fenglish:
myFT tooltip AB test for disengaged users was successful.
We are going to roll this out to 100% of users who fall into the predicted to disengage cohort.
However, the tooltip which encouraged users to visit myFT feed was not conclusive, we will not be rolling this out.

[Jira ticket](https://financialtimes.atlassian.net/browse/CON-130)

**Release** 👇 

Screenshot 2020-04-28 at 15 32 10

The tooltip is displayed for users who fill **All** the requirements below
- Users who are in disengaged cohort
- Users who haven't followed any topics
- Users who haven't seen the tooltip more than 3 times

**Kill** 👇 

Screenshot 2020-04-27 at 15 32 10

The tooltip was displayed for users who filled **All** the requirements below
- Users who are in disengaged cohort
- Users who visit articles via non-ft site
- Users who have followed concepts
- Users who haven't seen the tooltip more than 3 times

### DIFF
--- a/client/components/top/myft-disengaged-tooltip/main.js
+++ b/client/components/top/myft-disengaged-tooltip/main.js
@@ -1,18 +1,19 @@
+// This tooltip should be displayed for users who fill all the requirements below
+// - who are in disengaged cohort (it has been already checked when the decision for displaying this tooltip was made (messageSlotTop flag === 'MyftDisengagedTooltip'))
+// - who haven't followed any topics
+// - who haven't seen the tooltip more than 3 times
+
 import Tooltip from 'o-tooltip';
 import myftClient from 'next-myft-client/myft-bower';
 import {broadcast} from 'n-ui-foundations';
 import Cookies from 'js-cookie';
-import * as flags from '@financial-times/dotcom-ui-flags';
 
 const articleAddToMyftButton = document.querySelector('.topper__primary-theme .n-myft-follow-button');
 const ARTICLE_TOOLTIP_SEEN_COUNT_COOKIE_NAME = 'FT_MyFT_article_tooltip';
-const flagsClient = flags.init();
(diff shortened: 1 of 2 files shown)


client/components/top/myft-disengaged-tooltip/main.js
```diff
@@ -1,72 +1,33 @@
...
 
 const articleAddToMyftButton = document.querySelector('.topper__primary-theme .n-myft-follow-button');
-const headerMyFTLogo = document.querySelector('[data-trackable="my-ft"]');
-const externalReferer = document.referrer && !(new URL(document.referrer).hostname.endsWith('.ft.com'));
 const ARTICLE_TOOLTIP_SEEN_COUNT_COOKIE_NAME = 'FT_MyFT_article_tooltip';
-const {FT: {flags = {get: () => {}}} = {}} = window;
 let articleTooltipSeenCount = Cookies.get(ARTICLE_TOOLTIP_SEEN_COUNT_COOKIE_NAME) || 0;
 
 export default function customSetup (banner, done) {
-	if(!externalReferer && articleTooltipSeenCount >= 3) {
```
robsquires, line 14: Did we need to keep this `externalReferer` rule?

robsquires: As discussed IRL with @fenglish, it looks like this logic was moved up to prevent myft API calls that weren't required https://github.com/Financial-Times/n-messaging-client/pull/190.

By looking at the [original implementation](https://github.com/Financial-Times/n-messaging-client/pull/179) we can see`externalReferer` is relevant to the header tooltip only..so it's safe to be removed.

PR #285, opened 2020-05-27
ACQ-239 Use cookie to show or hide marketing popup prompt message

nickramsbottom:
Ammit doesn't use messaging data for unregistered users. Marketing Popup Prompt doesn't hide for 10 days after showing as a result. This PR uses a cookie with 10 day expiry to dismiss the notice instead of using Ammit.

## Advantges of this approach

The implementation is quite straight forward and doesn't add complexity to the Ammit code.

## Disadvantges of this approach

A user won't see lower priority messages instead of popup prompt when it is hidden because Ammit doesn't know that it is being hidden.

## Other solutions

Start of Ammit modification solution: https://github.com/Financial-Times/next-ammit-api/pull/671


emmalewis:
Not approving yet as this is in draft, but it all looks good!

PR #289, opened 2020-06-15
add accessibility text for cookie consent links 🐿 v2.13.0

alexflorisca:
This ads text to indicate to users that the "cookies" link on the cookie consent popup opens in a new window

### DIFF
--- a/server/templates/partials/bottom/cookie-consent.html
+++ b/server/templates/partials/bottom/cookie-consent.html
@@ -8,7 +8,7 @@
 	}}
 	<h2 class="cookie-banner__head" id="cookie-banner-aria-label">Cookies on FT Sites</h2>
 	<div class="cookie-banner__inner" data-o-banner-inner="" data-o-banner-inner="">
-		<div id="cookie-banner-aria-description">We use <a href="http://help.ft.com/help/legal-privacy/cookies/" class="cookie-banner__link cookie-banner__link--external" data-o-banner-action data-o-banner-action-type="view-policy" target="_blank" rel="noopener">cookies</a> for a number of reasons, such as keeping FT Sites reliable and secure, personalising content and ads, providing social media features and to analyse how our Sites are used.</div>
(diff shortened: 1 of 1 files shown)


server/templates/partials/bottom/cookie-consent.html
```diff
@@ -8,7 +8,7 @@
 	}}
 	<h2 class="cookie-banner__head" id="cookie-banner-aria-label">Cookies on FT Sites</h2>
 	<div class="cookie-banner__inner" data-o-banner-inner="" data-o-banner-inner="">
-		<div id="cookie-banner-aria-description">We use <a href="http://help.ft.com/help/legal-privacy/cookies/" class="cookie-banner__link cookie-banner__link--external" data-o-banner-action data-o-banner-action-type="view-policy" target="_blank" rel="noopener">cookies</a> for a number of reasons, such as keeping FT Sites reliable and secure, personalising content and ads, providing social media features and to analyse how our Sites are used.</div>
+		<div id="cookie-banner-aria-description">We use <a href="http://help.ft.com/help/legal-privacy/cookies/" class="cookie-banner__link cookie-banner__link--external" data-o-banner-action data-o-banner-action-type="view-policy" target="_blank" rel="noopener">cookies <span class="o-normalise-visually-hidden">opens in new window</span></a> for a number of reasons, such as keeping FT Sites reliable and secure, personalising content and ads, providing social media features and to analyse how our Sites are used.</div>
```
umbobabo, line 11: I think you should keep the `opens in new window` inside parenthesis cause I would expect a screen reader to do a pause after reading `cookies`. Maybe you can try it with VoiceOver in your machine and see the difference with or without.

I also believe the space should stay inside the span otherwise it will be visually rendered and it will sum with the space that is after the closure of the link, so you will have 2 spaces between `cookie` and `for a number of reasons`

```suggestion
		<div id="cookie-banner-aria-description">We use <a href="http://help.ft.com/help/legal-privacy/cookies/" class="cookie-banner__link cookie-banner__link--external" data-o-banner-action data-o-banner-action-type="view-policy" target="_blank" rel="noopener">cookies<span class="o-normalise-visually-hidden"> (opens in new window)</span></a> for a number of reasons, such as keeping FT Sites reliable and secure, personalising content and ads, providing social media features and to analyse how our Sites are used.</div>
```

alexflorisca: Thanks, good suggestions!

PR #290, opened 2020-06-16
ACQ-162 Create remaining articles bottom slot message

nickramsbottom:
Add a bottom slot message for users who are registered but don't have a subscription.

The message will only show (run `done`) on the client if `next-messaging-guru` returns a number of articles above zero.

Dynamic context is also added to tracking in this PR. This allows barrier guru results to be added to the tracking context when tracking events are triggered.

Tracking configuration coming in a future PR.

image

Related PR: https://github.com/Financial-Times/next-messaging-guru/pull/85


manifest.js
```diff
@@ -170,4 +170,11 @@ module.exports = {
 	tbybPostTrialSubscribe: {
 		path: 'bottom/tbyb-post-trial-subscribe'
 	},
+	friedFrankLicence: {
```
AniaMakes, line 173: any reason friedFrank message is still up? I think we took it down recently? Or is this a new one?

alexnaish: This doesn't necessarily mean the message is up, just that the code is still present

client/components/bottom/remaining-articles/main.scss
```diff
@@ -0,0 +1,25 @@
+.n-messaging-client-messaging-banner--remaining-articles {
+	text-align: center;
+	.o-banner__content {
+		p {
```
emmalewis, line 4: Could we use classes instead of styling the tag?

client/components/bottom/remaining-articles/main.scss
```diff
@@ -0,0 +1,25 @@
...
+
+			.remaining-articles__counter {
+				@include oTypographySerif(
+					$scale: 6,
+					$include-font-family: false,
+				);
+			}
+		}
+
+		a {
```
emmalewis, line 16: Could we use classes instead of styling the tag?

client/components/bottom/remaining-articles/main.scss
```diff
@@ -0,0 +1,25 @@
+.n-messaging-client-messaging-banner--remaining-articles {
+	text-align: center;
+	.o-banner__content {
+		p {
+			padding-bottom: oSpacingByName('s2');
+			@include oTypographySerif(3);
+
+			.remaining-articles__counter {
```
emmalewis, line 8: Is this nesting necessary?

client/next-messaging-guru-client.js
```diff
@@ -35,7 +35,8 @@ module.exports = function ({ guruEndpoint='/__message', name, guruQueryString }=
 			const validBannerThemes = [
 				'myft',
 				'product',
-				'marketing'
+				'marketing',
+				'slate-white'
```
emmalewis, line 39: Not a big deal at all but would `slate-background` or similar be more obvious here?

client/components/bottom/remaining-articles/main.js
```diff
@@ -0,0 +1,9 @@
+export default function customSetup (banner, done, guruResult) {
+	const remainingArticles = guruResult.renderData.remainingArticles;
+
+	if(remainingArticles === 0) {
```
emmalewis, line 4: will remaining articles always be defined? Would it be worth checking `if (!remainingArticles || remainingArticles === 0)`?

nickramsbottom:
@emmalewis changes made, thank you for reviewing!

client/components/bottom/remaining-articles/main.scss
```diff
@@ -0,0 +1,25 @@
+.n-messaging-client-messaging-banner--remaining-articles {
+	text-align: center;
+	.o-banner__content {
+		.remaining-articles__content {
```
emmalewis, line 4: Can these `.remaining-articles__*` classes just be root level, not nested at all?

nickramsbottom: yes!

PR #297, opened 2020-06-26
Energy source promo

jamesnicholls:
Adds message for Energy Source newsletter promo. See https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1128&modal=detail&selectedIssue=UG-192

Screenshot 2020-06-29 at 17 12 00


client/components/top/energy-source-promo/main.js
```diff
@@ -0,0 +1,68 @@
...
+				throw new Error(`Could not subscribe to newsletter ${response.statusText}`);
+			}
+
+			return true;
+		});
+};
+
+const setState = (banner, desiredState) => {
+	if (!['error', 'success', 'initial'].includes(desiredState)) {
+		return;
```
davidbailey00, line 28: should we log an error here?

client/components/top/energy-source-promo/main.js
```diff
@@ -0,0 +1,68 @@
...
+			contentEl.classList.remove('hidden');
+		} else {
+			contentEl.classList.add('hidden');
+		}
+	});
+};
+
+module.exports = function customSetup (banner, done) {
+	const submitButton = banner.messageElement.querySelector('[data-o-message-action="subscribe"]');
+	const closeButton = banner.messageElement.querySelector('[class*="__close"]');
```
davidbailey00, line 46: Does `__close` always come at the end? If so,

```suggestion
	const closeButton = banner.messageElement.querySelector('[class$="__close"]');
```

client/components/top/energy-source-promo/main.js
jamesnicholls: Yeah, we could throw an error that would then get logged by whatever app is using the component (hopefully).

client/components/top/energy-source-promo/main.js
davidbailey00: or could this be `.o-message__close` ?

jamesnicholls: Yeah that's probably best.

jamesnicholls:
> Nice! Would it be possible to have a screen grab in the description for if we want to reuse the design in the future?

I've added the screenshot. The design has a lot in common with the Coronavirus Business Update banner and one thing I considered was that if we start to use this design a lot, we could re-factor the templates and styling into a `newsletter-promo` component.

PR #301, opened 2020-07-02
Remove Envoy messages 🐿 v2.12.5

saldixon:
Removal of 28 messages which were used by Envoy (as per https://financialtimes.atlassian.net/browse/IDS-3):

appPromoMobile
b2bTrialAnon
b2bTrialContactUs
b2bTrialCountdown
b2bTrialMember
b2bTrialMobile
b2bTrialMyFt
b2bTrialNewsletter
b2cSubAcquisitions
b2cTrialAcquisitions
bookYourConsult
coronavirusNewsletterPromo
dailyDigest
fastFt
ftWeekendPromo
giftArticles
marketsData
myFtAccountTopicRecs
myFtFeedpage
myFtFeedpageOverview
nbeAutoSub
newsletterPromo
nextBestAction
onboardingMyFt
onboardingPremium
onboardingRegion
oneClickDailyDigest
techScrollAsia


manifest.js
```diff
@@ -51,116 +51,16 @@ module.exports = {
...
 		path: 'bottom/us-election-2020-promo'
 	},
-	coronavirusNewsletterPromo: {
-		path: 'bottom/coronavirus-newsletter-promo'
-	},
-	nextBestAction: {
-		path: 'top/next-best-action',
-		lazy: true
-	},
 	coronavirusNewsletterPromoTop: {
```
thurstontye, line 64: Looks like ` client/components/bottom/coronavirus-newsletter-promo/main.scss ` has been removed. So does this need to come out of the manifest or the style deletion reverted.

main.scss
```diff
@@ -2,21 +2,7 @@ $system-code: 'n-messaging-client';
 
 @import './client/themes';
 @import './client/components/top/anon-subscribe-now/main';
-@import './client/components/bottom/b2b-trial-myft/main';
```
thurstontye, line 5: Are these all solely driven by Envoy? It looks like in next-ammit at least `coronavirusNewsletterPromoTop` is in the messaging.json with no reference to envoy.

manifest.js
saldixon: that stylesheet was only used by the coronavirusNewsletterPromo message (deleted in this PR).  The remaining coronavirusNewsletterPromoTop message doesn't rely on its own bespoke stylesheet .. instead uses n-message* and o-message* classes to pick up styles.

main.scss
saldixon: The deleted ones are all as per instructions here:   https://financialtimes.atlassian.net/browse/IDS-3

coronavirusNewsletterPromoTop has never been controlled by Envoy.  (unlike coronavirusNewsletterPromo which was).

saldixon:
> Not 100% sure all of these are envoy messages.

Is it just the coronaVirus one(s) you're concerned about ?  It's confusing because there were two versions .. one was envoy controlled and has been removed, the other remains.  (controlled purely by Ammit)

manifest.js
thurstontye: ok

main.scss
thurstontye: Ok names too similar!

PR #302, opened 2020-07-03
Change conditional to hide message when less than 0 remaining 🐿 v2.13.0

nickramsbottom:
Users will see a final 'This is your last article' message before being shown the paywall.

This changes the message to show only when there are fewer than 0 articles left instead of fewer than 1.


nickramsbottom:
@ludrob great shout, I will add some docs in a second PR

PR #303, opened 2020-07-07
Improve the accessibility of the close button on banners - NOPS-257

GlynnPhillips:
The `close-button-label` option will be passed through to o-banner if a `closeLabel` parameter exists.

This will improve the accessibility of the close button as screen reader users will be given the context they need to decide if they want to close the banner.

Screenshot of o-banner markup including the extra close label

I have added a close label to each banner, I am new to most of these banners so I have tried to use my best judgement for the wording but I am open to better suggestions

### DIFF
--- a/server/templates/components/o-banner.html
+++ b/server/templates/components/o-banner.html
@@ -9,7 +9,8 @@
 	data-o-banner-banner-class="{{#if bannerClass}}{{bannerClass}}{{else}}o-banner{{/if}}"
 	data-o-banner-auto-open="false"
 	{{#if suppressCloseButton}} data-o-banner-suppress-close-button="true"{{/if}}
-	>
+	{{#if closeLabel}} data-o-banner-close-button-label="{{closeLabel}}"{{/if}}
+>
 	<div class="o-banner__outer">
 		{{#if @partial-block}}
 			{{{@partial-block}}}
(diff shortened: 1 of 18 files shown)


server/templates/components/o-banner.html
```diff
@@ -9,7 +9,8 @@
 	data-o-banner-banner-class="{{#if bannerClass}}{{bannerClass}}{{else}}o-banner{{/if}}"
 	data-o-banner-auto-open="false"
 	{{#if suppressCloseButton}} data-o-banner-suppress-close-button="true"{{/if}}
-	>
+	{{#if closeLabel}} data-o-banner-close-button-label="{{closeLabel}}"{{/if}}
```
nickcolley, line 12: Non-blocking: if we didn't guard this statement with the if statement, would it blow up if someone forgot to add a closeLabel? Could this be a good thing for it to be a requirement?

GlynnPhillips: Funny enough before lunch I started writing an issue on o-banner to propose making accessibility argument mandatory. Once I get the time to finish writing it I will tag you in it.

Regarding what would happen if we didn't guard it I think it would add empty attributes

GlynnPhillips:
Going to merge for now as I think this is an improvement but then see if anyone can check the copy and make iterations after if needed

PR #304, opened 2020-09-02
ACQ-342 Check whether message is lazy loaded when setting declarativeElement

nickramsbottom:
`n-messaging-client` is able to lazy load messages. This works for the bottom slot with examples in production. It looks like lazy loaded top slot messages haven't been used for a while and aren't working.

When trying to lazy load a message in the top slot nothing is rendered, only the placeholder template shows. This pull request fixes the client logic to check whether the message should be lazy loaded or not when deciding how to render using `o-message`. It also adds parameters passed to `o-message` to allow the message to load and attach to the correct parent element.

I imagine the logic can be simplified further to reflect how bottom slot messages are loaded. This is an initial fix.

### DIFF
--- a/client/top-slot.js
+++ b/client/top-slot.js
@@ -27,7 +27,6 @@ function getServerRenderedBanner (config, guruResult) {
 }
 
 module.exports = function ({ config={}, guruResult, customSetup }={}) {
-	let alertBanner;
 	const variant = (guruResult && guruResult.renderData && guruResult.renderData.dynamicTrackingData) || config.name;
 	const trackEventAction = config.name && generateMessageEvent({
 		flag: TOP_SLOT_FLAG,
@@ -36,19 +35,16 @@ module.exports = function ({ config={}, guruResult, customSetup }={}) {
 		trackingContext: config.trackingContext,
 		variant: variant
 	});
-	const declarativeElement = !config.lazy && getServerRenderedBanner(config, guruResult);
+	const declarativeElement = getServerRenderedBanner(config, guruResult);
 	const options = { messageClass: ALERT_BANNER_CLASS, autoOpen: false, close: message.getDataAttributes(declarativeElement).close};
 
-	if (declarativeElement) {
-		alertBanner = new message(declarativeElement, options);
-	} else if (guruResult && guruResult.renderData) {
-		alertBanner = new message(null, imperativeOptions(guruResult.renderData, options));
-	} else {
-		if (guruResult.skip && trackEventAction) {
-			trackEventAction('skip');
-		}
+	if (guruResult && guruResult.skip && trackEventAction) {
+		trackEventAction('skip');
 		return;
 	}
+
+	const allOptions = imperativeOptions(guruResult ? guruResult.renderData : {}, options);
+	const alertBanner = new message(declarativeElement, allOptions);
 
 	if (messageEventLimitsBreached(config.name)) {
 		trackEventAction('skip'); // todo do we actually need to do this?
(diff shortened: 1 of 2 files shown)


ststdimov (approved):

ludrob (approved):

client/top-slot.js
```diff
@@ -35,7 +36,7 @@ module.exports = function ({ config={}, guruResult, customSetup }={}) {
 		trackingContext: config.trackingContext,
 		variant: variant
 	});
-	const declarativeElement = getServerRenderedBanner(config, guruResult);
+	const declarativeElement = !config.lazy && getServerRenderedBanner(config, guruResult);
```
mifkata, line 39: The logic here is duplicate in `getServerRenderBanner` on line 14 and now this seems a bit invalid

mifkata (commented):
I'm not sure how to test this fully to know if the business logic is stable across multiple variants, so I'm just leaving some comments based on assumptions

client/top-slot.js
alexnaish: This looks like it'll nullify the middle condition of `getServerRenderedBanner` - the

```
if (config.lazy && guruResult && guruResult.renderData
			&& guruResult.renderData.reservedSpace) {				&& guruResult.renderData.reservedSpace) {
		// a hybrid of server-side and client-side solution:			// a hybrid of server-side and client-side solution:
		// this has reserved space on the page (to prevent "jerky" page loading)			// this has reserved space on the page (to prevent "jerky" page loading)
		// and we populate this space with lazily-fetched content			// and we populate this space with lazily-fetched content
		return config.content;
	}
```

client/top-slot.js
```diff
@@ -117,6 +118,8 @@ function imperativeOptions (opts, defaults) {
 			}
 		},
 		close: opts.closeButton,
-		dynamicTrackingData: opts.dynamicTrackingData
+		dynamicTrackingData: opts.dynamicTrackingData,
+		state: opts.state,
+		parentElement: TOP_SLOT_CONTENT_SELECTOR,
```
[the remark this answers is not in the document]
nickramsbottom: great spot! There's a decision to be made here. The bottom slot message has it set in n-messaging-client with messaging-guru not having a say.

@alexnaish @umbobabo do you have an opinion on which is preferable?

client/top-slot.js
nickramsbottom: Agreed. My plan is to get this to a point where it's working and a second refactor to have a look at the `getServerRenderBanner` function more closely to see what's going on.

I _think_ I just want line 22 to return a falsy value

client/top-slot.js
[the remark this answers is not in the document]
alexnaish: I would go with @mifkata's suggestion - gives us flexibility to override if necessary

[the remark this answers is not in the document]
nickramsbottom: I have reverted this for now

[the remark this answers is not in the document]
nickramsbottom: Done!

alexnaish (approved):

alexnaish (approved):

client/top-slot.js
```diff
@@ -99,8 +97,7 @@ function imperativeOptions (opts, defaults) {
 		autoOpen: opts.autoOpen || defaults.autoOpen,
 		messageClass: opts.messageClass || defaults.messageClass,
 		type: opts.type,
-		status: opts.status,
```
mifkata, line 102: does this needs to be removed as well?

nickramsbottom: yes! https://financialtimes.slack.com/archives/C02FU5ARJ/p1599121208020000

V3 of o-message used `status`, V4 uses `state`. This repo is on `^4.1.0`.

mifkata (approved):
Looks good! Thanks!

PR #305, opened 2020-09-03
ACQ-342 Implement intro offer message

nickramsbottom:
Ticket: https://financialtimes.atlassian.net/browse/ACQ-342
Design: https://marvelapp.com/prototype/24j3g43g/screen/72774236
Depends on & screenshots: https://github.com/Financial-Times/next-messaging-guru/pull/109


client/components/top/try-full-access/main.scss
```diff
@@ -0,0 +1,102 @@
...
+			}
+
+			.o-message__content-main {
+				display: flex;
+				flex-direction: column;
+
+				margin-top: oSpacingByName('s4');
+				margin-bottom: oSpacingByName('s6');
+				text-transform: uppercase;
+				font-size: 18px;
```
nickramsbottom, line 45: we might be able to use o-fonts for this: https://registry.origami.ft.com/components/o-fonts@4.4.1/

client/components/top/try-full-access/main.scss
```diff
@@ -0,0 +1,102 @@
...
+					margin-bottom: oSpacingByName('s8');
+					font-size: 24px;
+					line-height: 28px;
+				}
+			}
+		}
+
+		.o-message__content-detail {
+			color: oColorsByName('white');
+			.highlight {
```
nickramsbottom, line 60: is this rule covered by `o-message__content-highlight`?

manifest.js
```diff
@@ -79,5 +79,10 @@ module.exports = {
 	remainingArticlesTest: {
 		path: 'bottom/remaining-articles',
 		lazy: true
-	}
+	},
+	tryFullAccess: {
+		path: 'top/lazy',
+		lazy: true,
+		guruQueryString: 'offerId=683995bf-ed3f-2fd6-d589-1628ebad1629',
```
nickramsbottom, line 86: this needs to be the production offer ID:
```suggestion
		guruQueryString: 'offerId=41218b9e-c8ae-c934-43ad-71b13fcb4465',
```

client/components/top/try-full-access/main.scss
```diff
@@ -0,0 +1,102 @@
+@include oMessageAddState('try-full-access', (
+	'foreground-color': 'white',
```
mifkata, line 2: I think you need to use `oColorsByName` to be consistent with the rest of the codebase

nickramsbottom (commented):
Nice! Small changes

client/components/top/try-full-access/main.scss
nickramsbottom: good catch! examples here: https://github.com/Financial-Times/o-banner/blob/9579cf3211f7ac4b5506214e9a56ade949cbf7ee/demos/src/demo.scss#L28

valeksiev: The value provided is passed through `oColorsByName` by the [mixin here](https://github.com/Financial-Times/o-message/blob/f2caafb8475989b1f7664f0c300594809bec5211/src/scss/_state.scss#L19). It works both ways of course but is redundant. Do we want to be on the safe side?

Also in `n-messaging-client/client/components/o-mssage-notice.scss` (which I used as an example) string values are used. Do we unify when we see inconsistencies like this?

client/components/top/try-full-access/main.scss
valeksiev: I checked this page while working on this didn't find anything about the font size. Seems like `oFonts` is used only to load the font but the actual use is plain ol CSS.

Also in all other `scss` files in the project, the font-sizes are specified like this.

client/components/top/try-full-access/main.scss
valeksiev: I am not sure I understand the questions but this is what makes the first word of the second row pink. 😆

client/components/top/try-full-access/main.scss
```diff
@@ -0,0 +1,104 @@
...
+			@include oGridRespondTo(S) {
+				margin-left: 70px;
+			}
+			@include oGridRespondTo(M) {
+				margin-left: 132px;
+			}
+		}
+
+		.o-message__content-highlight {
+			color: oColorsByName('wheat');
```
nickramsbottom, line 77: can this be specified in `oMessageAddState`? line 4

client/components/top/try-full-access/main.scss
nickramsbottom: my comment on the other PR might make it clearer: https://github.com/Financial-Times/next-messaging-guru/pull/109#discussion_r490848138

client/components/top/try-full-access/main.scss
nickramsbottom: it's worth unifying, yes! good spot

client/components/top/try-full-access/main.scss
valeksiev: `oMessageAddState` updates both the highlight color and the button color so we need to override one of those. From the docs:
```
	'highlight-color': 'lemon', // lemon highlights with `o-message__content-highlight-color` and a lemon button
```

nickramsbottom: ahhh nice!

nickramsbottom (commented):
I can't approve my own PR but changes on top of mine look good!

mifkata (approved):
LGTM

PR #315, opened 2020-10-15
[ACQ-605] Add custom tracking event for top slot messages

mifkata:
# Overview
When a `top-slot` message is lazy loaded, its HTML is entirely rendered by the `oMessage` library, thus obscuring the DOM structure of the output. The original code depends on elements containing `[data-n-alert-banner-action]` data attribute and its value.

This code goes and sets that attribute onto the Primary action button, generated by `oMessage` and allows for normal code execution.


ststdimov (approved):
Not that much aware of how tracking works but seems logically correct as code changes. Also, I am OK with the PR from a code quality perspective

PR #336, opened 2021-01-26
Close the banner after a user has clicked the 'Leave Feedback'  button

bolajiadeyemi:
We need to close the banner after a user has clicked the 'Leave Feedback'  button.

### DIFF
--- a/client/components/bottom/post-activation-journey/main.js
+++ b/client/components/bottom/post-activation-journey/main.js
@@ -3,9 +3,9 @@ module.exports = (banner, done) => {
 	const closeButtons = Array.from(banner.bannerElement.querySelectorAll('.n-messaging-client-messaging-banner-post-activation-journey__action'));
 	closeButtons.forEach((closeButton) => {
 		closeButton.addEventListener('click', (event) => {
-			const isFeedbackBtn = event.target.name === LEAVE_FEEDBACK && typeof banner.bannerElement.remove === 'function';
+			const isFeedbackBtn = event.target.name === LEAVE_FEEDBACK;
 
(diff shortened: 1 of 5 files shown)


jamesnicholls (changes requested):
As discussed, you'll need to find a way of closing the pop-up when a customer clicks the "Leave feedback" button, but without triggering the `n-messaging:close` tracking event.

client/components/bottom/post-activation-journey/main.js
```diff
@@ -1,7 +1,13 @@
 module.exports = (banner, done) => {
-	const closeButton = banner.bannerElement.querySelector('.n-messaging-client-messaging-banner-post-activation-journey--close');
-	closeButton.addEventListener('click', function closePopup () {
-		banner.close();
+	const LEAVE_FEEDBACK = 'leave-feedback';
+	const closeButtons = Array.from(banner.bannerElement.querySelectorAll('.n-messaging-client-messaging-banner-post-activation-journey__action'));
+	closeButtons.forEach((closeButton) => {
+		closeButton.addEventListener('click', (event) => {
+			const isFeedbackBtn = event.target.name === LEAVE_FEEDBACK && typeof banner.bannerElement.remove === 'function';
```
jamesnicholls, line 6: **suggestion (non-blocking):** Could you use a different selector to target the "Leave feedback" button instead of using `name` here?

jamesnicholls (approved):
Tracking events are firing now, thanks. If you have a copy of Internet Explorer, can you test this in it please? IE doesn't support the `remove` method but it should be polyfilled by the [Polyfill service](https://polyfill.io/v3/)

PR #342, opened 2021-02-23
Update `o-cookie-message`

oliverturner:
This PR 

1. Pulls in a version of o-cookie-message that responds to backward navigation swipes and correctly displays / hides the Cookie Banner even when the page is reloaded from the [bfcache](https://web.dev/bfcache/)
1. Uses a relative link makes it easier to test user journeys locally
1. restores the redirect query parameter mentioned in [this discussion](https://github.com/Financial-Times/o-cookie-message/pull/145#discussion_r567738218) allowing users to navigate back to the page on which they were shown the cookie banner

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -44,6 +44,6 @@
     "n-myft-ui": "^22.0.0",
     "n-swg": "^2.1.1",
     "o-grid": "^5.2.6",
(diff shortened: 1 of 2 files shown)


alevito (approved):

alevito (approved):
:+1:

PR #343, opened 2021-02-23
Make manage cookies link more flexible and informative

oliverturner:
This PR makes two changes:
1. Using a relative link makes it easier to test user journeys locally
2. It restores the `redirect` query parameter mentioned in [this discussion](Financial-Times/o-cookie-message#145 (comment)) allowing users to navigate back to the page on which they were shown the cookie banner


alevito (approved):
Good job!

PR #344, opened 2021-02-24
Restore cookie message tracking

notlee:
I’ve refactored the bottom slot code some, mostly wrapping chunks in functions, to make sure I wasn’t missing anything. 

The refactoring is in a separate commit to actually adding tracking to o-cookie-message. It can be reviewed separately that way. Also if we want to throw that out to minimise the changes we’re making we can.

I've run through some paths to test manually, with cookie message and the post activation message. Here are cookie message events:
Screenshot 2021-02-24 at 11 45 49


client/bottom-slot.js
```diff
@@ -12,52 +12,52 @@ const BANNER_LINK_SELECTOR = `.${BANNER_CLASS}__link`;
...
-		// be revised after the major cascade
-		if (Array.isArray(guruResult.renderData.customThemes)) {
-			for (const theme of guruResult.renderData.customThemes) {
-				banner.bannerElement.classList.add(`${N_MESSAGING_BANNER_CLASS}--${theme}`);
-			}
-		}
+	// Setup banner tracking.
+	trackBannerInteractions(banner, trackEventAction);
 
+	// Display the banner.
```
oliverturner, line 43: question (non-blocking): Would it be clearer to test for the positive case?

client/bottom-slot.js
```diff
@@ -80,43 +80,52 @@ module.exports = function ({ config={}, guruResult, customSetup }={}) {
...
-		if (customSetup.default) {
-			customSetup = customSetup.default; // ESM modules
-		}
-		function customSetupCallback ({ skip=false }={}) {
-			if (skip) {
-				trackEventAction('skip');
-			} else {
-				banner.open();
-			}
+function renderBannerFromData (guruRenderData) {
```
oliverturner, line 85: Suggestion (non-blocking): a JSDoc comment documenting the types of `guruRenderData`'s properties could be helpful here

oliverturner (dismissed):
Couple of minor questions but approving as they're non-blocking.

client/bottom-slot.js
notlee: Debatably, I opted to test the negative case as its code block is short and easily parsed / skipped over whereas the positive case is quite the close block. Ideally I think `n-messaging-client` would be refactored so that each messaging component contains all its own logic to remove this conditional

client/bottom-slot.js
notlee: It would, I was limiting the time I needed to spend on this by omitting 🙈 I think `guruRenderData` is set by some other service ([next-messaging-guru](https://github.com/Financial-Times/next-messaging-guru)?) I don't know what values it may return without digging in, unless you know where its API is documented? I'd rather leave this to the maintainers to avoid an inaccurate type doc

client/bottom-slot.js
```diff
@@ -12,52 +12,52 @@ const BANNER_LINK_SELECTOR = `.${BANNER_CLASS}__link`;
...
-	const trackEventAction = config.name && generateMessageEvent({
-		flag: BOTTOM_SLOT_FLAG,
-		messageId: config.name,
-		position: config.slot,
-		trackingContext,
-		variant
-	});
+	const skip = guruResult && guruResult.skip;
+	const bannerLimitBreached = messageEventLimitsBreached(config.name);
+	if (skip | bannerLimitBreached) {
```
notlee, line 20:
```suggestion
	if (skip || bannerLimitBreached) {
```

replace bitwise OR with logical OR 😬 I think in this case the result is the same but whoops

oliverturner (approved):
Thank you!

alevito:
Code looks good!
I've also tested this locally on next-article and I can see all the Cookie banner events been sent to Spoor as expected.

alevito (approved):
:+1:

PR #360, opened 2021-04-01
Bower and npm compatible package

umbobabo:
This package forces team to use Bower, with this PR we introduce the necessary packages in NPM so consumers can use the library without Bower.

To be sure that npm packages are kept in sync with Bower packages a preinstall hook modify package.json with the Bower packages, if some of them do not have been mapped in the script it throws an error.

(documentation already stays that the package can be installed through npm / bower, so no updates in docs have to be done)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -41,7 +41,7 @@
     "o-overlay": "^3.0.0",
     "o-loading": "^4.0.0",
     "o-tracking": "^2.0.6",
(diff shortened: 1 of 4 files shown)


emortong (approved):
looks good! 

There are just some conflicts and failing snyk tests needed to be fixed before merging :)

emortong (approved):

PR #361, opened 2021-04-07
[ACQ-899] Add slot configuration to limit messages to be displayed

valeksiev:
What I've added to the read me is exactly the use case we have.

We want to add a messaging slot to a page but it should be able to display a subset of messages (just one in our case). This PR allows that. 

At first glance, the helper function deciding if the selected variant is one of the allowed is a bit of an overkill but because it is optional including the logic if it's provided, split it, then check it in the constructor will be actually harder.


ekocian (approved):
LGTM. not blocking: add some unit test to cover the logic

jonoft (dismissed):

valeksiev:
@ekocian I started adding tests but the presenter is written in such a way that testing it is quite hard. We already have examples for the wrapping function (the default export) and it's a lot more work to actually test the presenter class...

README.md
```diff
@@ -95,6 +95,23 @@ if ( window.FT.flags.messageSlotBottom || window.FT.flags.messageSlotTop ) {
...
 
+#### `variants` option
+
+If `variants` option is provided to the slot partial only the messages specified as a string of comma separated values will be displayed.
+For example:
+
+```
+{{> n-messaging-client/server/templates/slot type='bottom' variants='messageA,messageB'}}
+```
+would only display ONLY `messageA` or `messageB` if the "Brain™" decides one of those should be displayed. All other messages will be ignored.
```
maguero, line 106:
```suggestion
would display ONLY `messageA` or `messageB` if the "Brain™" decides one of those should be displayed. All other messages will be ignored.
```

maguero (approved):

README.md
ludrob, line 106:
```suggestion
would display ONLY `messageA` or `messageB` if the "Brain™" decides one of those should be displayed. All other messages will be ignored.
```

ludrob (dismissed):

ludrob:
> @ekocian I started adding tests but the presenter is written in such a way that testing it is quite hard. We already have examples for the wrapping function (the default export) and it's a lot more work to actually test the presenter class...

I was about to ask about tests as well, but that answers it.

ludrob (approved):

maguero (approved):

PR #369, opened 2021-05-19
changes cache keys to be branch-specific

leannecornish-ft:
currently the npm cache is for all branches. Nightly builds are broken after trying to use the cache saved by renovate-circleci-node-15.x. This PR is to split out the cache on a per branch basis.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,13 +22,13 @@ references:
 
   npm_cache_keys: &npm_cache_keys
     keys:
-        - v3-dependency-npm-{{ checksum "package.json" }}-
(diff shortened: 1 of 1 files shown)


apaleslimghost (approved):
ooh, nice find! is there a chance that we'll have the same problem on other repos that have had a Node v15 renovate PR?

leannecornish-ft:
> ooh, nice find! is there a chance that we'll have the same problem on other repos that have had a Node v15 renovate PR?

Possibly, depending on how they have their keys configured. If this is indeed the problem, I'll check all the others.

PR #374, opened 2021-06-21
UG-509 activation variant templates

Shamarni96:
Changed variant names from `onboarding` to `activation`


jamesnicholls (changes requested):
Good work. Everything works well. I have suggested some small improvements to the code. Could you check the design with Tasha/Dilyana?

Also could you delete the `package-lock.json` file? I don't think this repo uses it.

server/templates/partials/bottom/activation-not-complete.html
```diff
@@ -0,0 +1,25 @@
+<div class="n-messaging-client-messaging-banner n-messaging-client-messaging-banner--activation-not-complete">
+	<div class="o-banner o-banner--small o-banner--product" data-o-component="o-banner">
+		<div class="o-banner__outer">
+			<div class="o-banner__inner" data-o-banner-inner>
+			<div class="o-banner__content o-banner__content--long">
```
meel-io, line 5: **question**: why do we need `--long` and `--short` versions of the markup? I thought they were used if it was needed to have different versions by screen size but they look like they have the same content.

meel-io (commented):
The code looks good and it works perfectly on my local. I left a minor question but it's not a blocker.

bolajiadeyemi (dismissed):

meel-io (dismissed):

server/templates/partials/bottom/activation-not-complete.html
```diff
@@ -0,0 +1,19 @@
+<div class="n-messaging-client-messaging-banner n-messaging-client-messaging-banner--activation-not-complete">
+	<div class="o-banner o-banner--small o-banner--product" data-o-component="o-banner">
```
jamesnicholls, line 2: **suggestion:** You could `n-messaging-client/server/templates/components/o-banner` template to avoid a little bit of duplication here.

server/templates/partials/bottom/activation-not-complete.html
```diff
@@ -0,0 +1,19 @@
...
+				<h1>Discover the stories you will love</h1>
+				</header>
+				<p>Select the content that most interests you, from curated newsletters to podcasts and more.</p>
+			</div>
+			<div class="o-banner__actions">
+				<div class="o-banner__action">
+				<a href="https://www.ft.com/activate/bundles" class="o-banner__button">Explore now</a>
+				</div>
+			</div>
+			</div>
```
jamesnicholls, line 16: **nitpick:** The indentation is a little out-of-line here.

jamesnicholls (changes requested):
The client-side JS is not running because the `.o-banner` element is missing the `data-n-messaging-component`. I recommend that you use the `/server/templates/components/o-banner` template so that all the necessary attributes are in the HTML.

Also, the messages are missing the close button.

meel-io:
> The client-side JS is not running because the `.o-banner` element is missing the `data-n-messaging-component`. I recommend that you use the `/server/templates/components/o-banner` template so that all the necessary attributes are in the HTML.
> 
> Also, the messages are missing the close button.

Added the `/server/templates/components/o-banner` template and seems to be working fine 👍

bolajiadeyemi (approved):

jamesnicholls:
Tracking events are working now. Good to go!

leannecornish-ft (approved):

jamesnicholls (approved):

PR #379, opened 2021-08-23
ACC-1003 Add Top Banner for Single Term Renew message

AniaMakes:
# Description
Single Term users are user who are not on a subscription that auto renews. What this means, their subscription can expire. We want to encourage them to renew their subscription when that happens so we are adding a Top banner to FT pages. 

# Ticket
https://financialtimes.atlassian.net/browse/ACC-1003

# Screenshot in n-messaging-client demo
Yes, it looks a bit weird, but unless we release this and the changes in Ammit, this is the best I have for a screenshot
Screenshot 2021-08-23 at 14 26 35

# ... but I edited the NonPayment Banner to show the Renew Banner content, and they use the same component, so the Renew Banner should looks like so
**oops, the button should say Renew, but you get the idea**
Screenshot 2021-08-23 at 14 53 10
Screenshot 2021-08-23 at 14 53 25

### DIFF
--- a/manifest.js
+++ b/manifest.js
@@ -30,6 +30,9 @@ module.exports = {
 	paymentFailure: {
 		path: 'top/payment-failure'
 	},
+	singleTermRenew: {
+		path: 'top/renew-single-term-banner'
+	},
 	anonSubscribeNow: {
 		path: 'top/anon-subscribe-now',
 		trackingContext: {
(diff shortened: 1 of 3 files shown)


apaleslimghost (approved):
looks good!

binaryberry (approved):
Looks good to me!

PR #380, opened 2021-08-27
ACQ-1121: Static messages for Slider changes (Sept Campaign)

maguero:
### Description

`PR in DRAFT until Product team confirm the CTA action URLs`

Marketing September Campaign:
2 new versions of the slider:
- APAC region - XX% would read 33%
- All other regions - XX% would read 50%

Designs: https://www.figma.com/file/aEiiMtHu2qgOzHS118CzF4/Letters-to-this-New-World?node-id=9%3A18

How to test:
- Select the variant from: https://toggler.ft.com/#messageSlotBottom
image

- https://local.ft.com:5005/

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-1121

### Screenshots
| APC | ROW |
|---|---|
| image | image |
| image | image |


server/templates/partials/bottom/marketing-popup-prompt-sept-row.html
```diff
@@ -0,0 +1,33 @@
...
+			<p>Get Standard Digital access, including:</p>
+			<ul>
+				<li>Coronavirus Business Update: our guide to a disrupted world</li>
+				<li>Opinion &amp; analysis from leading business experts</li>
+				<li>Market-moving news, politics, tech, the arts and more</li>
+			</ul>
+		</div>
+		<div class="o-banner__content o-banner__content--short">
+			<header class="o-banner__heading n-messaging-banner__heading">
+				<h2 id="n-messaging-banner-aria-label">Special offer: Save 50%</h2>
```
angelaFT, line 21: Question: it seems to me that these two html partials are very similar - any chance of re-using them instead of having two? You've probably already thought about this...

ststdimov (approved):

server/templates/partials/bottom/marketing-popup-prompt-sept-row.html
maguero: You are right, they are almost the same, the only differences are the discount and the CTA link.
The reason it's "almost duplicated" it's because it doesn't worth to invest more time in a message for only 1 month, and we are removing them soon when the campaign ends.
Anyway its "reusing" `server/templates/components/o-banner` in some way 😄

angelaFT: Makes sense!

angelaFT (approved):

PR #392, opened 2021-09-22
UG-507: Show customer care message on customerCareCall variant

meel-io:
### Description
Added new template customer-care for bottom slot messages and assigned it to the activationNotStarted, activationNotComplete and activationSkipped variants replacing the current messages.

### Ticket
https://financialtimes.atlassian.net/browse/UG-507


jamesnicholls:
Looks good to me. The only problem I see is that the `n-messaging:act` event is failing, I think because the link goes to a page without `o-tracking` initialised. I think if we open the page in a new tab, it should fix the problem. I'm checking with Dilyana to see if opening the link in a new tab is okay.

jamesnicholls (approved):

leannecornish-ft (approved):

PR #401, opened 2021-10-27
ACC1241 - Open Day Premium Upsell banner

asugar13:
Open Day banner to show on premium articles for standard B2c users. This banner will be shown on Nov 3rd 2021.

Screenshot 2021-11-01 at 10 12 11


server/templates/partials/top/open-day-premium-upsell.html
```diff
@@ -0,0 +1,15 @@
+<div style="display:none" id="openDayBanner" class="o-message o-message--notice o-message--pikachu" data-o-component="o-message">
+	<div class="o-message__container">
+		<div class="o-message__content ">
+			<p class="o-message__content-main">
+				<span class="header-message">Enjoying this Premium article?</span> <br>
+					<span class="o-message__content-highlight">
+							READ MORE WITH A PREMIUM SUBSCRIPTION
+					</span> <br>
+					<span class="lower-message">Upgrade your subscription today for unrestricted access to all the FT has to offer across ft.com, the app and newsletters.</span>
```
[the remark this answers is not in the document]
asugar13: managed to change to p tags

client/components/top/open-day-premium-upsell/main.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.main-content {
+	display: inline-block;
+	margin-bottom: 20p;
+	vertical-align: middle;
+	padding-right: 100px;
+}
+
+a {
```
jamesnicholls, line 38: **issue:** I think this scope of this selector is too broad. It will affect any link on the page.

client/components/top/open-day-premium-upsell/main.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function customSetup (banner, done) {
+	let openDayBanner = document.getElementById('openDayPremiumUpsellBanner');
+
+	let target = document.querySelector('#site-content[data-access-level="premium"]');
+	if (target) {
+		openDayBanner.style.display = 'block';
```
jamesnicholls, line 6: **question:** Have you tested this in a screen reader (e.g. Mac OS Voiceover)? Do you know if the message will be announced by screen readers if it is displayed after the page has rendered?

server/templates/partials/top/open-day-premium-upsell.html
```diff
@@ -0,0 +1,16 @@
...
+		<div class="o-message__content">
+			<div class="main-content">
+				<p class="header-message">This Premium article and all others are free to read today</p>
+					<p class="o-message__content-highlight">
+						Want unrestricted access every day?
+					</p>
+					<p class="lower-message">Upgrade to a Premium Digital subscription to gain unlimited access to all we have to offer across ft.com, the app and newsletters.</p>
+			</div>
+			<div class="o-message__actions">
+					<a href="https://www.ft.com/manage/subscription/change/da448d11-851d-545c-dc0e-e9a2aef62e54" id="button" class="o-message__actions__primary">Upgrade to Premium</a>
```
jamesnicholls, line 12: **suggestion:** Any reason not to use `o-buttons` for the button element?

jamesnicholls: **issue:** I think you might have to give this element a class of `o-message__button` or `o-message_link` or a data attribute with the key `data-o-message-action` in order for the tracking events to work correctly. See https://github.com/Financial-Times/n-messaging-client/blob/72b9c0f8e5124afd1b5ef0308b86e0884ee07851/client/top-slot.js#L67-L94.

jamesnicholls (changes requested):
There are a couple of issues that I think need to be addressed before being deployed, like the overly broad `a` selector in the Sass file, and the ensuring the tracking events are being sent correctly.

Also I have a question about how displaying the message after the page has rendered affects accessibility.

client/components/top/open-day-premium-upsell/main.scss
asugar13: Understood and changing the selector to a class. I thought the styles would only be affected to the newly created component.

server/templates/partials/top/open-day-premium-upsell.html
asugar13: I have followed the origami example in https://www.ft.com/__origami/service/build/v3/demo?component=o-message%405.0.1&demo=notice-custom&system_code=origami-repo-data&brand=master
This was not using an `a` tag instead of `o-buttons` for the button component. However I can confirm that there are tracking events after clicking on the button. I believe it's because it's wrapped in a `div` with class `o-message__actions`

asugar13: Update: You were right. A generic click event was being generated but the `n-messaging-client` click event was not being generated. After some time on it I realized it wasn't only because of needing to have `data-o-message-action` on the `a` tag but also because I needed `data-n-messaging-component` on the top level div

client/components/top/open-day-premium-upsell/main.js
asugar13: I have tried this on Mac OS Voiceover and it worked fine (I can demo it to you if you want). I have added `aria-live="polite"` to the top level div just in case (I tried that on the Voiceover app too and it was fine as well)

server/templates/partials/top/open-day-premium-upsell.html
jamesnicholls: Hmm, I still don't see any `click` events being sent to Spoor.

asugar13: @jamesnicholls  please see the screen capture below. I have checked with @KirtiSiddapur from analytics

https://user-images.githubusercontent.com/10453619/139840542-701599b4-d30b-4f5d-9027-35fc31fc4847.mov

jamesnicholls (dismissed):

binaryberry (approved):
Looking great! The approach of relying on css tags to detect if an article is premium is fine given we're planning on reverting this in less than 48 hours, it doesn't have to be durable.
Well done for pulling this off, with the tracking to boot!

PR #409, opened 2021-11-08
UG-720 remove customerCareCall path

Shamarni96:
removed `customerCareCall` path from `manifest.js `. Also deleted the `customerCareCall` message template and `customer-care-call` directory from `client/components/bottom`


manifest.js
```diff
@@ -64,9 +64,6 @@ module.exports = {
 		path: 'bottom/remaining-articles',
 		lazy: true
 	},
-	customerCareCall: {
-		path: 'bottom/customer-care-call'
```
jamesnicholls, line 68: **chore:** Can you remove this template from the code base as well?

jamesnicholls (changes requested):
Thanks for the PR. Could you also remove the customer care call template and customer care component directory in `/client/components/bottom/`?

jamesnicholls (approved):

PR #411, opened 2021-11-22
CPP-642 migrate from circleci/* to cimg/*

serena97:
These images will be EOL from Dec 31st, and we should migrate to the modern cimg/* equivalents.


serena97:
automatically closed 🤖

.circleci/config.yml
```diff
@@ -6,12 +6,7 @@ references:
   container_config_node12: &container_config_node12
     working_directory: ~/project/build
     docker:
-      - image: circleci/node:12-browsers
-
-  container_config_lambda_node12: &container_config_lambda_node12
```
serena97, line 11: I removed this `container_config_lambda_node12` because it's not being used in the config.yml

leannecornish-ft (approved):
Looks good to me!

serena97:
FYI the old node-sass and n-foundations-ui version were using python2 and the new docker circleci img only had python3 not python2 so I updated these dependencies to a version that is using python3

PR #419, opened 2022-01-31
Bump o-tracking and next-myft-client

fenglish:


demos/src/demo.js
```diff
@@ -4,13 +4,8 @@ const { nMessagingClient } = require('../../main-client');
 document.documentElement.classList.add('js', 'enhanced');
 
 oTracking.init({
-	server: 'https://spoor-api.ft.com/px.gif',
```
fenglish, line 7: I delete this line because this is demo and no point to set fallback for no javascript users.
https://github.com/Financial-Times/o-tracking/tree/v3.1.4#tracking-without-javascript

PR #428, opened 2022-03-02
fix demos

apaleslimghost:
- switch node-sass to dart sass to fix `math` import
- fix import path for o-tooltip


maguero (approved):
Thank you very much for taking care of this! 🎉

PR #431, opened 2022-03-08
UG-823: Remove veil from embeddedOnboarding template when homepageOnboarding flag is not active

meel-io:
### Description

Removes veil set up by embeddedOnboarding template message when the homepageOnboarding flag is not active


jenniferemshepherd (dismissed):
😂

leannecornish-ft (approved):
LGTM 👍

PR #432, opened 2022-03-10
UG-729 remove postActivation journey now that the test is complete

jenniferemshepherd:
This test is finished and the variant needs removing.


meel-io (dismissed):
💯 sorry it took me so long to review 🇺🇸

PR #441, opened 2022-04-01
UG-842: Remove dangling tag from embeddedOnboarding template

meel-io:
### Description

Removes an unclosed div tag on the message template that was breaking the layout of the message's consumer.


jamesnicholls (approved):

jenniferemshepherd (approved):
tbf none of the reviewers spotted it either ;)

PR #444, opened 2022-04-11
CPP-809 Document v9 branch split and backporting workflow

ivomurrell:


ivomurrell:
I think it should be okay staying in the migration section – it doesn't affect people who are just starting to use the project as they won't be using Bower

PR #452, opened 2022-04-27
UG-730/UG-731 Remove tbybInTrial and tbybPostTrial subscribe messages

Shamarni96:
removed `tbybInTrialSubscribe` and `tbybPostTrialSubscribe` messages and removed both variants from the `messageSlotBottom` flag.


leannecornish-ft (commented):
Looks good from our side 👍 
Nitpick: `tbybInTrialSubscribe` and `tbybPostTrialSubscribe` are still being referenced in an example in corporate-cancellation-message.md - might be worth removing so as not to be confusing?

leannecornish-ft (approved):
looks good to me!

serena97 (approved):

PR #453, opened 2022-04-28
ACC1627 - fix typo Danske Bank message

asugar13:
fix a typo for accesibility purposes in the Danske Bank message.


danieleruiz (approved):
good point

PR #457, opened 2022-05-18
UG-885 Implement config, styles and behaviour for Academia sharing message

jamesnicholls:
Adds config, styles, and client-side behaviour for the Academia sharing message. The message is lazy, and retrieves it's HTML structure from next-messaging-guru (see https://github.com/Financial-Times/next-messaging-guru/pull/178).

In order to test:
1. Set the `messageSlotBottom` flag to the variant `AcademiaCTA`
2. Set the `academicCtaTest` flag to `variant`
3. Checkout the associated [next-messaging-guru PR](https://github.com/Financial-Times/next-messaging-guru/pull/178) and run using `make run`
4. Checkout this PR and run the app using `make demo-with-guru`
5. Go to https://local.ft.com:5005/
6. The message should appear as per [the designs](https://www.figma.com/file/09mlE9M0XVG5fIZy99tls2/(Final)-Professor-Engagement)


serena97 (dismissed):

leannecornish-ft (commented):
Looks good 👍  I just have one question - on mobile view, the copy link button is now quite large as the url has changed, just checking we are ok with that from a style perspective?
Screenshot 2022-06-10 at 18 53 22

jamesnicholls:
> Looks good 👍 I just have one question - on mobile view, the copy link button is now quite large as the url has changed, just checking we are ok with that from a style perspective?

Yeah, good question. I had a concern about that before I left on holiday but forgot to check with Tasha and/or Dilyana. Will check with them now. Thanks for the review.

leannecornish-ft (approved):
thanks James, looks good to me 👍

client/components/bottom/academia-cta/main.scss
```diff
@@ -77,6 +77,10 @@
 			background-origin: border-box;
 			text-decoration: none;
 			padding-right: 20px;
+			-webkit-box-orient: vertical;
```
serena97, line 80: question: are we sure we want to use `-webkit-box-orient` here? the docs https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient mentions that it has been deprecated and to use flexbox instead

jamesnicholls: Oh good spot. Unfortunately, `-webkit-line-clamp` doesn't work with `display:flex`, so I'll have to find a different approach.

serena97 (approved):

PR #459, opened 2022-05-25
ACC-1691 Revert "Acc1627 Add Danske Bank Cancellation Message for v9 release"

Nazehs:
Reverts Financial-Times/n-messaging-client#455

Reverting Danske Bank cancellation message 

Task  jira link
https://financialtimes.atlassian.net/browse/ACC-1691


asugar13 (approved):
approving!

PR #463, opened 2022-05-26
ACC1691 remove n-myft-ui from devDependencies

asugar13:
[JIRA](https://financialtimes.atlassian.net/browse/ACC-1691)


Nazehs (approved):
approved

PR #466, opened 2022-06-21
Update x-follow-button to v8

debugwand:
while I cannot find any references to x-follow-button within the code, I am updating the x-dash components just in case there is some downstream usage of it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@financial-times/x-follow-button": "1.0.4",
+        "@financial-times/x-follow-button": "8.0.2",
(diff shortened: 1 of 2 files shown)


ivomurrell:
**suggestion**: seeing as bumping the dependency by 7 major versions (albeit with no breaking changes that would obviously affect `x-follow-button`) could possibly break any packages relying on it transitively anyway, I say we should just get rid of this dependency if we aren't using it directly

debugwand:
I had checked what the changes are and there aren't any, because we release all the packages under each new version. However, I did some digging through the history and found that there was once a direct usage of it, and this wasn't tidied up properly. Therefore, I'm removing. If something was relying on it, it shouldn't have been.

ivomurrell (approved):

PR #485, opened 2023-01-18
chore: remove unused o-share dependency

frshwtr:
A breaking release of [o-share](https://github.com/Financial-Times/origami/blob/main/components/o-share/CHANGELOG.md#900-2023-01-17) was created recently, resulting in conflicting dependencies.

After reviewing the code, it appears `o-share` is not used and can be removed.


ivomurrell (approved):
That makes things easier, thanks!

PR #487, opened 2023-02-06
ug-1103 add support for n-myft-ui v30 peer dep

leannecornish-ft:
N-myft-ui is now on version 30, so this PR is to allow it as a peer dependency.


leannecornish-ft:
Closing as there's another PR to remove the dependency alogether: https://github.com/Financial-Times/n-messaging-client/pull/488

PR #500, opened 2024-06-26
Remove polyfill.io reference

rowanmanning:
Boring and know it's not a risk but let's clean up.


joelcarr (approved):
Agreed. As this is a docs change - no need to release a new version as far as I can tell. Do you agree?

